PR #7, opened 2020-09-07
Implement swap and burn

cloudhead:
Add swap & burn on registration. Uses uniswap as the price oracle and exchange.

### DIFF
--- a/contracts/DummyRouter.sol
+++ b/contracts/DummyRouter.sol
@@ -18,11 +18,15 @@ contract DummyRouter is Router {
         address to,
         uint256
     ) external override payable returns (uint256[] memory) {
-        rad.transfer(to, amountOutMin);
+        require(msg.value > 0, "Swap amount should be positive");
+
+        uint256 tokenAmount = msg.value;
+
+        rad.transfer(to, tokenAmount);
 
         uint256[] memory amounts = new uint256[](2);
         amounts[0] = msg.value;
-        amounts[1] = amountOutMin;
+        amounts[1] = tokenAmount;
 
         return amounts;
     }
--- a/contracts/DummyRouter.sol
+++ b/contracts/DummyRouter.sol
@@ -1,0 +1,33 @@
+// spdx-license-identifier: gpl-3.0-only
+// solhint-disable func-name-mixedcase
+pragma solidity ^0.6.12;
+
+import "./Rad.sol";
+import "./Router.sol";
+
+contract DummyRouter is Router {
+    Rad private rad;
+
+    constructor(address _rad) public {
+        rad = Rad(_rad);
+    }
+
+    function swapExactETHForTokens(
+        uint256 amountOutMin,
+        address[] calldata,
+        address to,
+        uint256
+    ) external override payable returns (uint256[] memory) {
+        rad.transfer(to, amountOutMin);
+
+        uint256[] memory amounts = new uint256[](2);
+        amounts[0] = msg.value;
+        amounts[1] = amountOutMin;
+
+        return amounts;
+    }
+
+    function WETH() external override pure returns (address) {
+        return address(0x123456789abcdef);
+    }
+}
--- a/contracts/DummyRouter.sol
+++ b/contracts/DummyRouter.sol
@@ -28,6 +28,11 @@ contract DummyRouter is Router {
         amounts[0] = msg.value;
         amounts[1] = tokenAmount;
 
+        require(
+            tokenAmount >= amountOutMin,
+            "Token amount swapped should be greater than minimum"
+        );
+
         return amounts;
     }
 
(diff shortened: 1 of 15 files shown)


contracts/FixedWindowOracle.sol
```diff
@@ -0,0 +1,87 @@
...
+            uint32 blockTimestamp
+        ) = Uniswap.currentCumulativePrices(address(pair));
+
+        uint32 timeElapsed = blockTimestamp - blockTimestampLast; // overflow is desired
+
+        // ensure that at least one full period has passed since the last update
+        require(
+            timeElapsed >= PERIOD,
+            "ExampleOracleSimple: PERIOD_NOT_ELAPSED"
+        );
```
cloudhead, line 67: Will change this to not abort.

contracts/libraries/Uniswap.sol
```diff
@@ -0,0 +1,91 @@
+// Nb. The code in this file was adapted from the Uniswap libraries.
```
cloudhead, line 1: This is copied from Uniswap, no need to review.

contracts/FixedWindowOracle.sol
```diff
@@ -0,0 +1,83 @@
+// SPDX-License-Identifier: GPL-3.0-only
+// Nb. The code in this file was taken & adapted from Uniswap.
```
cloudhead, line 2: This is also mostly taken from Uniswap. I've simplified a couple things only.

contracts/FixedWindowOracle.sol
```diff
@@ -0,0 +1,83 @@
...
+
+import "@uniswap/v2-core/contracts/interfaces/IUniswapV2Factory.sol";
+import "@uniswap/v2-core/contracts/interfaces/IUniswapV2Pair.sol";
+import "@uniswap/lib/contracts/libraries/FixedPoint.sol";
+
+import "./libraries/Uniswap.sol";
+
+// fixed window oracle that recomputes the average price for the entire period once every period
+// note that the price average is only guaranteed to be over at least 1 period, but may be over a longer period
+contract FixedWindowOracle {
```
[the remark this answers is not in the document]
lftherios: +1 on using a proper uniswap pair

contracts/DummyRouter.sol
```diff
@@ -0,0 +1,37 @@
...
+    ) external override payable returns (uint256[] memory) {
+        require(msg.value > 0, "Swap amount should be positive");
+
+        uint256 tokenAmount = msg.value;
+
+        rad.transfer(to, tokenAmount);
+
+        uint256[] memory amounts = new uint256[](2);
+        amounts[0] = msg.value;
+        amounts[1] = tokenAmount;
```
CodeSandwich, line 29: We're returning a 2-element dynamic array. Could we get away with a tuple? It should be cheaper and easier to work with.

contracts/DummyRouter.sol
```diff
@@ -0,0 +1,37 @@
...
+
+        return amounts;
```
CodeSandwich, line 31: We're returning a 2-element dynamic array. Could we get away with a tuple? It should be cheaper and easier to work with.

cloudhead: We can't because we're using tring to be compatible with Uniswap's router, which uses these signatures. I'll document that.

contracts/Exchange.sol
```diff
@@ -0,0 +1,45 @@
...
+
+        address[] memory path = new address[](2);
+        path[0] = router.WETH();
+        path[1] = address(rad);
+
+        // The minimum amount of Rad we're willing to receive in exchange.
+        uint256 minimumRad = oracle.consultEthRad(msg.value);
+
+        uint256[] memory amounts = router.swapExactETHForTokens(
+            minimumRad,
```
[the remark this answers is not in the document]
cloudhead: You're right. Fixed in https://github.com/radicle-dev/radicle-contracts/pull/7/commits/f53ac355528864bedfadd9b578420b14804aa94e

contracts/FixedWindowOracle.sol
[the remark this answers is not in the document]
cloudhead: This is very complicated to setup, but I agree we should somehow figure out how to do it.

contracts/Exchange.sol
```diff
@@ -0,0 +1,45 @@
...
+            path,
+            receiver,
+            block.timestamp
+        );
+
+        uint256 radAmount = amounts[1];
+        require(radAmount >= minimumRad, "Enough Rad was exchanged");
```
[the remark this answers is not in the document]
cloudhead: Yes, but when deployed, it will be the uniswap contract that handles this call, so we're just making sure it's not broken. However since I changed the minimum to be zero, this is no longer relevant anyway :+1:

contracts/DummyRouter.sol
cloudhead: https://github.com/radicle-dev/radicle-contracts/pull/7/commits/119bb70992854b831233975618593e34157b33cc

geigerzaehler (approved):

PR #13, opened 2020-09-11
Deploy uniswap/ens backed contracts

cloudhead:
This sets up a deploy function, `deployAll` that deploys the bytecode of ENS and Uniswap, sets up proper liquidity pairs for Uniswap and hooks everything up with our contract. This allows us to test everything using the "real" contracts, instead of the dummy ones.

As you'll notice, the registration fee taken from the Uniswap price oracle is always going to be slightly lower than expected, eg. given a 1:1 liquidity ratio, the price of 1000 USD will be ~999 ETH.

* [x] Deploy Uniswap-based exchange
* [x] Deploy ENS-based registrar

### DIFF
--- a/contracts/Exchange.sol
+++ b/contracts/Exchange.sol
@@ -6,9 +6,9 @@ import "./PriceOracle.sol";
 import "./Router.sol";
 
 contract Exchange {
-    Router private router;
-    PriceOracle private oracle;
-    address private rad;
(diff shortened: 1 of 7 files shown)


lftherios (approved):

NunoAlexandre (approved):
solid :ok_hand: awesome to test the real thing.

PR #14, opened 2020-09-17
Add ReceiverWeights list

CodeSandwich:
Adds a data structure, which is needed for the proper pool implementation (PR coming soon). I have a few questions in code comments.

### DIFF
--- a/contracts/Pool.sol
+++ b/contracts/Pool.sol
@@ -28,3 +28,79 @@ contract Pool {
         receiver.transfer(value);
     }
 }
+
+struct ReceiverWeight {
+    address next;
+    uint32 weight;
+}
+
+/// @notice Helper methods for receiver weights list.
+/// The list works optimally if after applying a series of changes it's iterated over.
+/// The list uses 1 word of storage per receiver with a non-zero weight.
+library ReceiverWeightsImpl {
+    address internal constant ADDR_ROOT = address(0);
+    address internal constant ADDR_UNINITIALIZED = address(0);
+    address internal constant ADDR_END = address(1);
+
+    /// @notice Return the next non-zero receiver weight and its address.
+    /// Removes all the zeroed items found between the current and the next receivers.
+    /// Iterating over the whole list removes all the zeroed items.
+    /// @param current The previously returned receiver address or zero to start iterating
+    /// @return next The next receiver address
+    /// @return weight The next receiver weight, zero if the end of the list was reached
+    function nextWeight(
+        mapping(address => ReceiverWeight) storage self,
+        address current
+    ) internal returns (address next, uint32 weight) {
+        next = self[current].next;
+        weight = 0;
+        if (next != ADDR_END && next != ADDR_UNINITIALIZED) {
+            weight = self[next].weight;
+            // remove elements being zero
+            if (weight == 0) {
+                do {
+                    address newNext = self[next].next;
+                    // Somehow it's ~1500 gas cheaper than `delete self[next]`
+                    self[next].next = ADDR_UNINITIALIZED;
+                    next = newNext;
+                    if (next == ADDR_END) break;
+                    weight = self[next].weight;
+                } while (weight == 0);
+                // link the previous non-zero element with the next non-zero element
+                // or ADDR_END if it became the last element on the list
+                self[current].next = next;
+            }
+        }
+    }
+
+    /// @notice Get weight for a specific receiver
+    /// @param receiver The receiver to get weight
+    /// @return weight The receinver weight
+    function getWeight(
+        mapping(address => ReceiverWeight) storage self,
+        address receiver
+    ) internal view returns (uint32 weight) {
+        weight = self[receiver].weight;
+    }
+
+    /// @notice Set weight for a specific receiver
+    /// @param receiver The receiver to set weight
+    /// @param weight The weight to set
+    /// @return previousWeight The previously set weight, may be zero
+    function setWeight(
+        mapping(address => ReceiverWeight) storage self,
+        address receiver,
+        uint32 weight
+    ) internal returns (uint32 previousWeight) {
+        previousWeight = self[receiver].weight;
+        self[receiver].weight = weight;
+        // Item not attached to the list
+        if (self[receiver].next == ADDR_UNINITIALIZED) {
+            address rootNext = self[ADDR_ROOT].next;
+            self[ADDR_ROOT].next = receiver;
+            // The first item ever added to the list, root item not initialized yet
+            if (rootNext == ADDR_UNINITIALIZED) rootNext = ADDR_END;
+            self[receiver].next = rootNext;
+        }
+    }
+}
--- a/contracts/PoolTest.sol
+++ b/contracts/PoolTest.sol
@@ -1,0 +1,49 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.6.2;
+pragma experimental ABIEncoderV2;
+
+import "./Pool.sol";
+
+contract ReceiverWeightsTest {
+    using ReceiverWeightsImpl for mapping(address => ReceiverWeight);
+
+    mapping(address => ReceiverWeight) private receiverWeights;
+    ReceiverWeightIterated[] private receiverWeightsIterated;
+    int256 public receiverWeightsSumDelta;
+
+    struct ReceiverWeightIterated {
+        address receiver;
+        uint32 weight;
+    }
+
+    function setWeights(ReceiverWeightIterated[] calldata weights) external {
+        int256 weightsSumDelta = 0;
+        for (uint256 i = 0; i < weights.length; i++) {
+            ReceiverWeightIterated calldata weight = weights[i];
+            weightsSumDelta += weight.weight;
+            weightsSumDelta -= receiverWeights.setWeight(
+                weight.receiver,
+                weight.weight
+            );
+        }
+        receiverWeightsSumDelta = weightsSumDelta;
+        delete receiverWeightsIterated;
+        address receiver = address(0);
+        while (true) {
+            uint32 weight;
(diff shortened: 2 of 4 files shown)


contracts/Pool.sol
```diff
@@ -28,3 +28,80 @@ contract Pool {
...
+
+struct ReceiverWeight {
+    address next;
+    uint32 weight;
+    bool isAttached;
+}
+
+/// @notice Helper methods for receiver weights list.
+/// The list works optimally if after applying a series of changes it's iterated over.
+/// The list uses 1 word of storage per receiver with a non-zero weight.
```
CodeSandwich, line 40: Based on Solidity docs I think that it's true, but I'm not sure how to prove that.

contracts/Pool.sol
```diff
@@ -28,3 +28,80 @@ contract Pool {
...
+            weight = nextReceiverWeight.weight;
+            // remove elements being zero
+            if (weight == 0) {
+                do {
+                    next = nextReceiverWeight.next;
+                    // the end of the list is reached
+                    if (next == address(0)) break;
+                    // delete nextReceiverWeight to release its memory
+                    nextReceiverWeight.isAttached = false;
+                    nextReceiverWeight.next = address(0);
```
CodeSandwich, line 67: Is there a better solution? `delete nextReceiverWeigth;` fails to compile, because Solidity thinks that I want to clear the pointer and make it invalid. I don't know how to tell it that I want to delete the pointee.

contracts/PoolTest.sol
```diff
@@ -0,0 +1,48 @@
...
+        uint32 weight;
+    }
+
+    function setWeights(ReceiverWeightIterated[] calldata weights) external {
+        int256 weightsSumDelta = 0;
+        for (uint256 i = 0; i < weights.length; i++) {
+            weightsSumDelta += weights[i].weight;
+            weightsSumDelta -= receiverWeights.setWeight(
+                weights[i].receiver,
+                weights[i].weight
```
CodeSandwich, line 25: It hurts my eyes, but `ReceiverWeightIterated calldata weight = weights[i];` does not compile. It's not possible to store a structure on the stack either. Is the only way to clean up the code here to copy the interesting item to memory? That would be so wasteful!

contracts/Pool.sol
```diff
@@ -28,3 +28,84 @@ contract Pool {
...
+        weight = 0;
+        if (next != ADDR_END) {
+            ReceiverWeight storage nextReceiverWeight = self[next];
+            weight = nextReceiverWeight.weight;
+            // remove elements being zero
+            if (weight == 0) {
+                do {
+                    next = nextReceiverWeight.next;
+                    // delete nextReceiverWeight to release its memory
+                    nextReceiverWeight.next = ADDR_UNINITIALIZED;
```
CodeSandwich, line 70: Is there a better solution? `delete nextReceiverWeigth;` fails to compile, because Solidity thinks that I want to clear the pointer and make it invalid. I don't know how to tell it that I want to delete the pointee.

contracts/Pool.sol
```diff
@@ -28,3 +28,84 @@ contract Pool {
...
+    /// Removes all the zeroed items found between the current and the next receivers.
+    /// Iterating over the whole list removes all the zeroed items.
+    /// @param current The previously returned receiver address or zero to start iterating
+    /// @return next The next receiver address
+    /// @return weight The next receiver weight, zero if the end of the list was reached
+    function nextWeight(
+        mapping(address => ReceiverWeight) storage self,
+        address current
+    ) internal returns (address next, uint32 weight) {
+        ReceiverWeight storage currentReceiverWeight = self[current];
```
CodeSandwich, line 59: I couldn't find much information about the optimizer.

Would it cost anything if I replaced every usage of `currentReceiverWeight` with `self[current]`? Key hashing cost is real, but does the optimizer prevent its recalculation?

contracts/Pool.sol
```diff
@@ -28,3 +28,84 @@ contract Pool {
...
+        next = currentReceiverWeight.next;
+        weight = 0;
+        if (next != ADDR_END) {
+            ReceiverWeight storage nextReceiverWeight = self[next];
+            weight = nextReceiverWeight.weight;
```
CodeSandwich, line 64: Does it make sense to copy the storage value to the stack just to read it? If I used `nextReceiverWeight.weight` everywhere, would EVM actually access storage every time? Or would the optimizer make it read only once and reuse this value from stack?

contracts/Pool.sol
cloudhead: It should be easy to run the tests and see how much gas is used in either case.

contracts/PoolTest.sol
```diff
@@ -0,0 +1,48 @@
...
+        }
+    }
+
+    function getReceiverWeightsIterated()
+        external
+        view
+        returns (ReceiverWeightIterated[] memory)
+    {
+        return receiverWeightsIterated;
+    }
```
cloudhead, line 47: if `receiverWeightsIterated` is `public`, it can be accessed directly via `contract.receiverWeightsIterated()`

contracts/Pool.sol
```diff
@@ -28,3 +28,83 @@ contract Pool {
...
+        if (receiverWeight.next == ADDR_UNINITIALIZED) {
+            ReceiverWeight storage rootReceiverWeight = self[ADDR_ROOT];
+            address rootNext = rootReceiverWeight.next;
+            rootReceiverWeight.next = receiver;
+            // The first item ever added to the list, root item not initialized yet
+            if (rootNext == ADDR_UNINITIALIZED) rootNext = ADDR_END;
+            receiverWeight.next = rootNext;
+        }
+    }
+}
```
cloudhead, line 110: So this is basically implemented as a linked list. I worry that this is expensive and complex. Can we not use a flat array?

CodeSandwich: Could you explain? How a flat array would be cheaper to manage?

contracts/PoolTest.sol
CodeSandwich: Not quite, `contract.receiverWeightsIterated()` requires an index parameter and returns a single item. That's inconvenient for testing.

contracts/Pool.sol
CodeSandwich: I was looking in ethers.io docs for functionality like this, but I couldn't find it, do you have a tip up your sleeve? If not, I'll look harder.

contracts/Pool.sol
CodeSandwich: Bypassed

contracts/Pool.sol
CodeSandwich: It's optimized

contracts/Pool.sol
CodeSandwich: It doesn't matter, it's optimized

contracts/Pool.sol
cloudhead: `TransactionReceipt::gasUsed()` will give you the amount. We can use that to optimize later.

contracts/PoolTest.sol
cloudhead: Ah gotcha.

contracts/Pool.sol
cloudhead: A linked list forces us to keep an extra `next` pointer per item, and a lookup for every iteration. I'm wondering if traversing a flat array would be cheaper and result in simpler code as well :man_shrugging:

contracts/Pool.sol
```diff
@@ -28,3 +28,79 @@ contract Pool {
...
+    address internal constant ADDR_END = address(1);
+
+    /// @notice Return the next non-zero receiver weight and its address.
+    /// Removes all the zeroed items found between the current and the next receivers.
+    /// Iterating over the whole list removes all the zeroed items.
+    /// @param current The previously returned receiver address or zero to start iterating
+    /// @return next The next receiver address
+    /// @return weight The next receiver weight, zero if the end of the list was reached
+    function nextWeight(
+        mapping(address => ReceiverWeight) storage self,
```
cloudhead, line 52: How come this is passed into each function, vs. being part of the contract global scope?

cloudhead (approved):
I think this is good to merge. I don't quite understand the rationale of some of the functionality, eg. why does `nextWeight` delete all zero weights, but I'm sure I'll get a clearer picture when the rest of the code is there.

contracts/Pool.sol
CodeSandwich: Iterating is easier, but modifications become way harder. How to update a weight for address `X`? How to tell update from creation of a new item? How to delete an item? Each of these operations would require iteration over the whole or at least an arbitrary part of the list. The mapping-linked list hybrid solves both iteration and modification use cases.

contracts/Pool.sol
CodeSandwich: I don't think that it has any gas cost at all, but it lets us have more than one weight mapping. It'll be needed for delegation pools. It also gives us better logic encapsulation.

CodeSandwich:
> why does nextWeight delete all zero weights

Good question! This is the cheapest way to prune storage memory. If we wanted to clear memory as soon as an item is deleted, we would need access to both the previous and the next items. The list would have to be a double linked list, which would double storage usage. Because we're pruning storage while iterating, we know what was the previous element, so the list can be single linked. The list is going to be iterated after every pool update anyway, so storage pruning at that moment is enough to keep it perfectly tight.

contracts/PoolTest.sol
CodeSandwich: Completely false, it just works. VSCode's in-editor error highlighter is wrong.

contracts/Pool.sol
CodeSandwich: Thanks :)

PR #22, opened 2020-09-30
Add 'prepare' script so package can be installed as git dep

geigerzaehler:
With this change the package can be installed as a git dependency with Yarn. See https://github.com/yarnpkg/yarn/issues/5235#issuecomment-571206092

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
   },
   "scripts": {
     "build": "buidler compile && tsc",
+    "prepare": "yarn prepack",
     "prepack": "yarn build",
     "test": "buidler test",
     "lint": "yarn run lint:prettier:check && yarn run lint:solhint && yarn run lint:eslint",


.npmignore
```diff
@@ -0,0 +1,6 @@
+/artifacts
+/build
```
NunoAlexandre, line 2: don't we need the `build` dir to **not** be ignored so that yarn can pick it up?

geigerzaehler: You’re right.

NunoAlexandre: Let's give this a spin locally to make sure it works before merging.

package.json
```diff
@@ -30,6 +30,7 @@
   },
   "scripts": {
     "build": "buidler compile && tsc",
+    "prepare": "yarn prepack",
```
NunoAlexandre, line 33: Who will be calling `yarn prepare`? Is it automatically the first option git dependencies if provided?

.npmignore
geigerzaehler: It was working all along. Turns out that the `.npmignore` file is not needed because we specifiy `files` explicitly in `package.json`.

package.json
geigerzaehler: Yarn will be calling it when installing this as a dependency in another project.

.npmignore
NunoAlexandre: even better :+1:

NunoAlexandre (approved):
:flying_disc:

PR #23, opened 2020-09-30
Fail on lint warnings

geigerzaehler:
Also enforces function type signatures


CodeSandwich (approved):
This is so cool, I was trying to fix these warnings, but at some point I gave up

PR #28, opened 2020-10-01
Implement simple attestations

### LINKED ISSUE #20: User Attestations
To verify that a user of radicle's funding contracts is the same user as a user on the radicle-link level, we need some kind of proof that is verifiable by any client, given a connection to both Ethereum and the Radicle Link network.

The use-cases are the following:
* Given the maintainer of a radicle link project, find their Ethereum donation address.
* Given an Ethereum address, find out if it belongs to any user of radicle link.

Basically, we'd like any client (eg. upstream) to easily relate and link Ethereum addresses with radicle link (stable) identifiers.

The proposed method only requires state on Ethereum.

---

**Proposal**

To create the link between the two networks, the user creates an **attestation** on chain (Ethereum). This attestation is essentially a signature of the radicle link id, stored in the Ethereum state and associated with the user's Ethereum account, by means of the transaction author (`msg.sender` in solidity).

```solidity
// Mapping of all attestations.
mapping(address => Attestation) attestations;

struct Attestation {
  bytes32 id; // The radicle-link id (root hash)
  bytes64 signature; // signature using one of the device keys (certifiers) under the radicle identity
  bytes32 publicKey; // the public (device) key associated with the above signature
}
```

The signature is created off-chain like so:

```
signature = secretKey.sign(ethereumAccountId + radicleId)
```

And stored on chain, using the transaction author as the associated ethereum account. Eg.

```solidity
function attest(bytes32 id, bytes64 signature, bytes32 publicKey) {
  attestations[msg.sender] = Attestation { id, signature, publicKey };
}
```

To verify the attestation:

1. Get the latest set of public keys under the radicle link identity being verified.
2. Get the attestation from the Ethereum state. (Eg. using the radicle link id via the reverse index)
3. Verify that the public key in the attestation is one of the keys authorized to sign on behalf of the identity in (1)
4. Verify the attestation signature using the public key in the attestation, and a concatination of the radicle id with the ethereum account id. ie. `public_key.verify(attestation.signature, &[ethereum_account_id, attestation.id])`

---

If the device key used to attest is lost, a new attestation should be made. This way, the device key can be revoked with no problem.

Note also that with this method, the identity doc on radicle link doesn't need to be updated when an attestation is made.

cloudhead:
Closes #20 

A simple attestation scheme. These attestations need to be verified off-chain, by verifying the signature. In the radicle-link identity doc of an attested user, we should store the ethereum address that was used to attest.


contracts/AttestationRegistry.sol
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.6.12;
+
+contract AttestationRegistry {
+    struct Attestation {
+        bytes32 id; // Holds a 256-bit hash
```
kim, line 6: This means that `id` is essentially opaque, is that correct?

I think that's ok, but keep in mind that the verifier needs to know both the "root hash" of the `link` identity (the thing in the URN) _and_ the revision which introduced the claim of the address.

test/attestations.test.ts
```diff
@@ -0,0 +1,96 @@
...
+    const [signer] = await ethers.getSigners();
+    const address = await signer.getAddress();
+    const attestationRegistry = await new AttestationRegistryFactory(
+      signer
+    ).deploy();
+    await attestationRegistry.deployed();
+
+    const id = ethers.utils.randomBytes(32);
+    const pk = ethers.utils.randomBytes(32);
+    const sig = [
```
geigerzaehler, line 17:
```suggestion
    const sig = new Array(64).fill([0])
```

contracts/AttestationRegistry.sol
```diff
@@ -0,0 +1,31 @@
...
+        bytes32 publicKey; // Holds an ed25519 key
+        bytes1[64] signature; // Holds an ed25519 signature
```
geigerzaehler, line 8: Alternativley

```suggestion
        byte[64] signature; // Holds an ed25519 signature
```

Reads slightly better in my opinion.

test/attestations.test.ts
cloudhead: Was hoping someone would help me out here, I spent 30 mins trying to unify the types.

contracts/AttestationRegistry.sol
cloudhead: Agreed but I think prettier did that. Happy to change the lint though if you know what it is..

contracts/AttestationRegistry.sol
cloudhead: Ah interesting! 32 bytes won't be enough then. Should we include a revision number? (cheap) Or a revision hash? (more expensive)

kim: Ya it would cost 2x20 bytes for as long as git uses SHA-1.

It's a design decision -- being able to reify the revision from the chain state may simplify verification, because you can content-address it. If you have the `link` history already, you can also compute the hash (if we think of `id` as `HASH(root || revision)`) by looking for the revision which introduced the claim. If, otoh, you want to construct a `link` URN from the chain, we need the `root` to be stored separately.

cloudhead: Cool, makes sense. Let's see how expensive it actually is : )

contracts/AttestationRegistry.sol
cloudhead: Confirmed, it's `prettier` :unamused:

geigerzaehler: Ok. Then let’s keep it

contracts/AttestationRegistry.sol
cloudhead: In my tests, using `bytes20` vs `bytes32` was actually more expensive :thinking: -- so I stuck to `bytes32` as this also gives us future-compat when git switches to sha2

CodeSandwich (approved):

contracts/AttestationRegistry.sol
geigerzaehler: > In my tests, using `bytes20` vs `bytes32` was actually more expensive

Using `bytes20` does not reduce the storage cost because every field has to be stored in its separate 32 byte storage slot. It becomes more expensive because now we have to zero-extend 20 bytes to 32 bytes.

geigerzaehler (approved):

contracts/AttestationRegistry.sol
```diff
@@ -0,0 +1,34 @@
...
+        bytes1[64] calldata signature
+    ) public {
+        attestations[msg.sender].id = id;
+        attestations[msg.sender].revision = revision;
+        attestations[msg.sender].publicKey = publicKey;
+        attestations[msg.sender].signature = signature;
+    }
+
+    /// Revoke any attestation made by the sender.
+    function revokeAttestation() public {
```
kim, line 31: One thing to meditate over is if we'd want to provide a way to issue a _signed_ revocation. That is, if I still have access to a valid `link` key, migrating to a new identity would be verifiably legit. Doing so may or may not make sense, though

contracts/AttestationRegistry.sol
cloudhead: Good to know!

contracts/AttestationRegistry.sol
cloudhead: Ah that's a good point. I think both options will probably be needed, in case you lose your *link* key. Clients would have to handle them differently though :thinking:

PR #33, opened 2020-10-05
Add and verify dependencies for web3 bindings

geigerzaehler:
We add the necessary dependencies to use the Web3 contract bindings. We also ensure that TypeScript checks the bindings.


CodeSandwich (approved):
Inching toward Web3 :+1:

PR #35, opened 2020-10-06
Increase solidity line width to 100

CodeSandwich:


cloudhead (approved):
This helps function signatures a lot :+1:

Issue #21, opened 2020-09-29
Orgs


cloudhead:
The idea behind the Orgs feature is to keep a canonical list of projects under control of an entity, whether it be a DAO, a user or a set of users. The umbrella under which these projects are kept is called an "Org". Some key features:

* An org has a unique ENS-compatible name, eg. `monadic.radicle.eth`.
* Orgs are controlled by an Ethereum address. This can be a regular address, or a smart contract address.
* Projects can be registered under an org.
* Projects registered under an org can have their state (or commit) hash anchored on-chain.
* Orgs have an account balance.
* Clients (eg. Upstream) should display Orgs as a list of projects, each with their latest anchored hash.

Orgs can be implemented by deploying an `Org` contract for each org.
```
contract Org {  
  struct Project {
    bytes32 id; // radicle link id
    bytes32 hash; // HEAD
  }

  Project[] projects; // Projects under this org
  address owner; // Org admin user/contract
  mapping(bytes32 => uint) index; // Project index

  constructor() {
    owner = msg.sender;
  }

  function anchorProject(bytes32 id, bytes32 hash) {
    require(msg.sender == owner);
    projects[index[id]].hash = hash;
  }
}
```

cloudhead:
PTAL @lftherios @CodeSandwich @geigerzaehler

CodeSandwich:
Should it be possible to have a project ID and based on it find all the head hashes and corresponding org/users? It sounds like a reasonable discovery step when deciding who to follow.

cloudhead:
That's not a usecase I'm aware of.. I would bundle that more with "search" which will have to be based on off-chain data anyway. I think the flow we're imaginging is that orgs can be found by their **name**, once they are registered, and then you can find projects under them. But there is no need to find all instances of a given project.

The reason for having an org contract instance per org is so that we can give orgs their own accounts, which I think is a nice feature.

CodeSandwich:
Ok, if that's not the feature, then IMO the design is good and simple :+1:

juliendonck:
> * Projects can be registered under multiple orgs.

@cloudhead What's the use case for this? I remember chatting about this, but forgot where we landed. I would be worried that people would register other's projects to get the donations to them.

cloudhead:
[...]

PR #38, opened 2020-10-07
Implement basic orgs

cloudhead:
Addresses parts of #21 

This changeset doesn't implement the full org functionality; only project anchoring. The rest will come in subsequent changesets.

An org contract is meant to be deployed per org. This allows orgs to have an address, and makes the contract a little simpler.

### DIFF
--- a/contracts/Org.sol
+++ b/contracts/Org.sol
@@ -1,0 +1,60 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.6.12;
+
+contract Org {
+    /// A project anchor under an org.
+    struct Project {
+        bytes32 rev; // Project revision hash
+        bytes32 hash; // Project state (eg. HEAD commit) hash
+    }
+
+    /// Org owner/admin.
+    address public owner;
+
+    /// Index of projects under this org, keyed by project identifier.
+    /// The values represent offsets into the `projects` array.
+    mapping(bytes32 => Project) public projects;
+
+    modifier ownerOnly {
+        require(msg.sender == owner, "Only the org owner can perform this action");
+        _;
+    }
+
+    /// Construct a new org instance, by providing an owner address.
+    constructor(address _owner) public {
+        owner = _owner;
+    }
+
+    /// Check whether a project exists.
+    function projectExists(bytes32 id) public returns (bool) {
+        return projects[id].hash != bytes32(0);
+    }
+
+    // ~ All functions below should use the `ownerOnly` modifier ~
+
+    /// Set the org owner.
+    function setOwner(address _owner) public ownerOnly {
+        owner = _owner;
+    }
+
+    /// Anchor a project state on chain, by providing a revision and hash. This method
+    /// should be used for adding new projects to the org, as well as updating
+    /// existing ones.
+    function anchorProject(
+        bytes32 id,
+        bytes32 rev,
+        bytes32 hash
+    ) public ownerOnly {
+        require(hash != bytes32(0), "The project hash must not be the zero hash");
+
+        Project storage proj = projects[id];
+
+        proj.rev = rev;
+        proj.hash = hash;
+    }
+
+    /// Remove a project from the org.
+    function removeProject(bytes32 id) public ownerOnly {
+        delete projects[id];
+    }
+}
--- a/contracts/Org.sol
+++ b/contracts/Org.sol
@@ -15,11 +15,6 @@ contract Org {
     /// The values represent offsets into the `projects` array.
     mapping(bytes32 => Project) public projects;
 
-    modifier ownerOnly {
-        require(msg.sender == owner, "Only the org owner can perform this action");
-        _;
-    }
-
     /// Construct a new org instance, by providing an owner address.
     constructor(address _owner) public {
         owner = _owner;
@@ -31,6 +26,12 @@ contract Org {
     }
 
     // ~ All functions below should use the `ownerOnly` modifier ~
+
+    /// Functions that can only be called by the contract owner.
+    modifier ownerOnly {
+        require(msg.sender == owner, "Only the org owner can perform this action");
+        _;
+    }
 
     /// Set the org owner.
     function setOwner(address _owner) public ownerOnly {
--- a/contracts/Org.sol
+++ b/contracts/Org.sol
@@ -21,7 +21,7 @@ contract Org {
     }
 
     /// Check whether a project exists.
-    function projectExists(bytes32 id) public returns (bool) {
+    function projectExists(bytes32 id) public view returns (bool) {
         return projects[id].hash != bytes32(0);
     }
 
--- a/test/org.test.ts
+++ b/test/org.test.ts
@@ -69,7 +69,9 @@ describe("Org", function () {
     // Should revert!
     await submit(org.connect(bob).anchorProject(id, rev, hash))
       .then(() => expect.fail("Expected error"))
-      .catch(() => {});
+      .catch(() => {
+        /* OK */
+      });
 
     // Ok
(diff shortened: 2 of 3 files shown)


test/org.test.ts
```diff
@@ -0,0 +1,129 @@
...
+    await org.deployed();
+
+    const id = ethers.utils.randomBytes(32);
+    const rev = ethers.utils.randomBytes(32);
+    const hash = ethers.utils.randomBytes(32);
+
+    // Should revert!
+    await submit(org.connect(bob).anchorProject(id, rev, hash))
+      .then(() => expect.fail("Expected error"))
+      .catch(() => {});
```
cloudhead, line 72: Not sure this is the best way to expect failure, but couldn't get other ways to work.

contracts/Org.sol
```diff
@@ -0,0 +1,61 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.6.12;
+
+contract Org {
```
CodeSandwich, line 4: The design here is to have a separate contract instance per org. It's rather expensive and non-trivial to use. It's also impossible to introduce any updating mechanism, the contract is ultimately set in stone. Would it be possible to have a single instance of the contract? Is having a fully customized instance our desired use case?

test/org.test.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    // Ok
+    await submit(org.connect(owner).setOwner(bobAddr));
+
+    // Should revert!
+    await submit(org.connect(owner).setOwner(ownerAddr))
+      .then(() => expect.fail("Expected error"))
+      .catch(() => {
+        /* OK */
+      });
```
CodeSandwich, line 109: This looks like a perfect material for a tool in `support.ts`. Iff it's possible to crash a test from inside of an `async` function, I remember having some problems with it.

CodeSandwich (commented):
Nice and simple :+1:

test/org.test.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    // Ok
+    await submit(org.connect(owner).anchorProject(id, rev, hash));
+
+    // Should revert!
+    await submit(org.connect(bob).removeProject(id))
+      .then(() => expect.fail("Expected error"))
+      .catch(() => {
+        /* OK */
+      });
```
geigerzaehler, line 84: Unfortunately this does not work as expected. The exception thrown by `expect.fail()` will be caught `.catch()` and we will not see any test failures. I’d suggest the following, which can easily be extracted into a helper:
```suggestion
    const receipt = await (await org.connect(bob).removeProject(id).wait())
    assert.equal(receipt.status, 0)
```

contracts/Org.sol
cloudhead: I *think* this is the right design.

* For upgrades, it's simply a matter of pointing the Org's ENS name to a new contract. The big difference is that it's not a global upgrade of all org contracts. This has pros and cons.
* Having a contract per org gives orgs extra flexibility. It's theoretically possible to publish a different contract per org with the same ABI.
* Orgs having their own accounts is, I think, quite important. I'm a little skeptical of a single contract managing the money of *all* orgs for example. I would be worried about this from a security point of view.
* I don't really see it being expensive: yes, the initial deployment is more expensive, but once that's out of the way, it should be slightly cheaper to do day-to-day operations (eg. anchoring), since the contract is simpler. This should amortize over time :thinking: 
* The non-trivial aspect I guess is that users need to deploy this contract. However they won't have to do that manually: this is something that we should handle in an "OrgFactory" contract, that deploys the "Org" contract for you. This gives us a bunch of flexibility, and allows us to change the default org contract that is deployed for *new* deployments, as well as charge or reimburse the user as needed. This pattern is very common with multi-sig wallets.
* On the other hand, the Org contract is easier to audit and get right with this design, then an "OrgRegistry" contract that has to handle all orgs.

It's hard to know exactly if this decision will be the right one today, but I tend to prefer this model to the "global" contract model for the reasons above.

CodeSandwich: I can't argue with any of that, it makes perfect sense.

So most importantly we're providing an API, the implementation is just a help for the casual, lazy users. I'm not sure what's the convention, should we publish a separate `.sol` file with just the interface, so others can easily import it and then implement or use it?

test/org.test.ts
cloudhead: Hmmm, I thought I tested this and it worked. The reason I didn't go for the solution you proposed is that I thought it wouldn't work due to the exception being raised before we can check the status.

contracts/Org.sol
cloudhead: Yeah that's a good point. I think in the beginning, 99% of users will just use this contract, but eventually we might want to standardize on an interface. I would say let's open an issue for that. I don't want to prematurely define the interface, but I think we should have it set in stone for the official launch.

cloudhead: #41

test/org.test.ts
cloudhead: Ok you're right though about it being swallowed :man_facepalming:

cloudhead: Seems like this works: adaaca1

test/org.test.ts
cloudhead: adaaca1

CodeSandwich (approved):

test/support.ts
```diff
@@ -20,6 +20,27 @@ export async function submit(
...
+  let succeeded = false;
+
+  try {
+    const receipt = await (await tx).wait();
+
+    if (receipt.status == 1) {
+      succeeded = true;
+    }
+  } catch {
+    /* OK */
```
geigerzaehler, line 36: I don’t think we should ignore exceptions here. If one is raised this means that the transaction was not included in a block which is probably not what we want. If you remove the `try/catch` block then the tests should still pass but we will also be alerted when something is wrong with submitting the transcation.

CodeSandwich: But it's inside `submitFailing`, it's supposed to revert

cloudhead: If I remove the try/catch, the test will fail, because an uncaught exception will be raised.

test/support.ts
```diff
@@ -20,6 +20,27 @@ export async function submit(
...
+  }
+
+  if (succeeded) {
+    expect.fail("Expected transaction to fail");
+  }
```
CodeSandwich, line 41: BTW how about this? Transaction failure always throws anyway.
```suggestion
  try {
    await (await tx).wait();
  } catch {
    return;
  }
  expect.fail("Expected transaction to fail");
```

test/support.ts
```diff
@@ -20,6 +20,27 @@ export async function submit(
   return receipt;
```
CodeSandwich, line 20: Also here `assert.equal(receipt.status, 1, "transaction must be successful");` is dead code, it can't be ever triggered.

test/support.ts
cloudhead: Yeah, cleaner.

test/support.ts
cloudhead: What do you mean? The code runs as long as the tx doesn’t throw

CodeSandwich: Yes, but if a tx doesn't throw, this assertion won't throw either. When a tx unexpectedly reverts, the test never prints `transaction must be successful`, an exception is thrown earlier.

cloudhead: I see so you’re saying the status is always ‘1’ in cases where it doesn’t throw.

CodeSandwich: Exactly

geigerzaehler (approved):

Issue #34, opened 2020-10-06
Get all recipients and funds per block of my pool


geigerzaehler:
In upstream we want to show all recipients of a user’s pool and the funds per block that the pool disburses. For this the contract should expose a view that returns at least all recipient addresses. As a follow-up we would also need the recipient weights.

Issue #29, opened 2020-10-02
Add getter for pool sender's receivers


CodeSandwich:
This is almost surely necessary for the MVP UI

PR #40, opened 2020-10-08
Get all recipients and funds per block of my pool 

CodeSandwich:
Fixes https://github.com/radicle-dev/radicle-contracts/issues/34 and https://github.com/radicle-dev/radicle-contracts/issues/29

### DIFF
--- a/contracts/Pool.sol
+++ b/contracts/Pool.sol
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: GPL-3.0-only
 pragma solidity ^0.6.2;
+pragma experimental ABIEncoderV2;
 
 /// @notice Funding pool contract. Automatically sends funds to a configurable set of receivers.
 ///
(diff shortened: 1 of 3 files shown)


contracts/Pool.sol
```diff
@@ -80,6 +81,11 @@ contract Pool {
         mapping(uint64 => int256) amtDeltas;
     }
 
+    struct ReceiverWeight {
+        address receiver;
+        uint32 weight;
+    }
```
geigerzaehler, line 87: Suggestions:
```suggestion
    struct Receiver {
        address address;
        uint32 weight;
    }
```

contracts/Pool.sol
```diff
@@ -176,6 +182,16 @@ contract Pool {
         senders[msg.sender].amtPerBlock = uint192(amount);
     }
 
+    /// @notice Gets the target amount sent on every block from the sender of the message.
+    /// On every block this amount is rounded down to the closest multiple of the sum of the weights
```
geigerzaehler, line 186:
```suggestion
    /// The actual amount send on every block differs from the target value. It is is rounded down to the closest multiple of the sum of the weights
```

contracts/Pool.sol
CodeSandwich: Good point, but `struct Receiver` is already defined on line 73. It's used to store the complete information about a receiver and I think that it fits this name better than a container for an address-weight pair.

contracts/Pool.sol

geigerzaehler (approved):

PR #42, opened 2020-10-13
Pool test

CodeSandwich:
Improve test coverage for pool contract and fix bugs, which were found. In the next PR the test utilities will be reused across the whole repository.

### DIFF
--- a/contracts/Pool.sol
+++ b/contracts/Pool.sol
@@ -208,7 +208,7 @@ contract Pool {
         require(sender.weightSum <= SENDER_WEIGHTS_SUM_MAX, "Too much total receivers weight");
         if (weight != 0 && oldWeight == 0) {
             sender.weightCount++;
-            require(sender.weightSum <= SENDER_WEIGHTS_COUNT_MAX, "Too many receivers");
+            require(sender.weightCount <= SENDER_WEIGHTS_COUNT_MAX, "Too many receivers");
         } else if (weight == 0 && oldWeight != 0) {
             sender.weightCount--;
         }
@@ -416,6 +416,7 @@ library ReceiverWeightsImpl {
         address receiver,
         uint32 weight
     ) internal returns (uint32 previousWeight) {
+        require(receiver > address(1), "Invalid receiver address");
         previousWeight = self.data[receiver].weight;
         self.data[receiver].weight = weight;
         // Item not attached to the list
--- a/test/pool.test.ts
+++ b/test/pool.test.ts
@@ -5,11 +5,28 @@ import {
 import {Pool} from "../contract-bindings/ethers/Pool";
 import {ReceiverWeightsTest} from "../contract-bindings/ethers/ReceiverWeightsTest";
 import buidler from "@nomiclabs/buidler";
-import {Signer} from "ethers";
+import {Signer, BigNumber, BigNumberish, ethers} from "ethers";
 import {assert, expect} from "chai";
 
 async function addr(idx: number): Promise<string> {
   return await (await buidler.ethers.getSigners())[idx].getAddress();
+}
+
+function randomAddress(): string {
+  return numberToAddress(ethers.utils.randomBytes(20));
+}
+
+function numberToAddress(num: BigNumberish): string {
+  const hex = ethers.utils.hexlify(num);
+  const padded = ethers.utils.hexZeroPad(hex, 20);
+  return ethers.utils.getAddress(padded);
+}
+
+function getRevertCause(error: Error): string {
+  return error.message.replace(
+    "VM Exception while processing transaction: revert ",
+    ""
+  );
 }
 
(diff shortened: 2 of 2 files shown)


test/pool.test.ts
```diff
@@ -31,6 +37,27 @@ async function callOnNextBlock<T>(fn: () => T): Promise<T> {
...
+  );
+  return Promise.all(poolSigners);
+}
+
+interface PoolSigner {
+  pool: Pool;
+  addr: string;
+}
+
+async function newPoolSigner(pool: Pool, signer: Signer): Promise<PoolSigner> {
```
geigerzaehler, line 54: Only used once. I’d prefer to inline it.

test/pool.test.ts
```diff
@@ -31,6 +37,27 @@ async function callOnNextBlock<T>(fn: () => T): Promise<T> {
...
+async function getPoolSigners(): Promise<PoolSigner[]> {
+  const signers = await buidler.ethers.getSigners();
+  const pool = await deployPool(signers[0]);
+  const poolSigners = signers.map(async (signer: Signer) =>
+    newPoolSigner(pool, signer)
+  );
+  return Promise.all(poolSigners);
+}
+
+interface PoolSigner {
```
geigerzaehler, line 49: How about calling it `User`? I’m not happy with the “Signer” part because it has no relation to `Signer`.

test/pool.test.ts
```diff
@@ -5,13 +5,19 @@ import {
...
 import buidler from "@nomiclabs/buidler";
-import {Signer} from "ethers";
+import {Signer, BigNumber, BigNumberish, ethers} from "ethers";
 import {assert, expect} from "chai";
 
 async function addr(idx: number): Promise<string> {
   return await (await buidler.ethers.getSigners())[idx].getAddress();
 }
 
+function numberToAddress(num: number): string {
```
geigerzaehler, line 15: We could call this `generateAddress()` and randomly generate an address instead of using the `num` parameter. We don’t rely on this to be deterministic and if we need to reuse it we can always save the address in a local variable

test/pool.test.ts
```diff
@@ -80,12 +107,17 @@ async function setReceiver(
...
   }
-  expectReceivers(pool, receivers);
+  await expectReceivers(pool, receivers);
 }
 
 async function getAllReceivers(pool: Pool): Promise<Map<string, number>> {
   const receivers = await pool.getAllReceivers();
-  return new Map(receivers.map(({receiver, weight}) => [receiver, weight]));
+  return new Map(
+    receivers.map(({receiver, weight}: {receiver: string; weight: number}) => [
```
geigerzaehler, line 116: Do we need an explicit type annotation? This seemed to be ok before.

test/pool.test.ts
```diff
@@ -116,6 +148,19 @@ async function withdraw(
...
 }
 
+async function expectSetReceiverReverts(
+  pool: Pool,
+  address: string,
+  weight: number
+): Promise<void> {
+  try {
+    await pool.setReceiver(address, weight);
+  } catch {
```
geigerzaehler, line 158: We should probably inspect the error to ensure that we’re only catching the correct ones and re-throw others.

contracts/Pool.sol
```diff
@@ -416,6 +416,7 @@ library ReceiverWeightsImpl {
         address receiver,
         uint32 weight
     ) internal returns (uint32 previousWeight) {
+        require(receiver > address(1), "Invalid receiver address");
```
geigerzaehler, line 419: Why isn’t this `receiver != address(0)`?

geigerzaehler (commented):
I’ve reviewed the infratstrcuture. Looking at the tests now

geigerzaehler:
I have no additional comments about the tests. They seem pretty exhaustive :clap:

contracts/Pool.sol
CodeSandwich: Because `ADDR_END = address(1)` and setting a weight to this address is an error too

test/pool.test.ts
CodeSandwich: I was sure that it was producing warnings, but you're right, it's unnecessary, fixed

test/pool.test.ts
CodeSandwich: We do, we generate explicit addresses for `0` and `1`. Random addresses are indeed a little cleaner and easier to use, added `randomAddress` and switched to it wherever possible.

test/pool.test.ts
CodeSandwich: Perfect name! Fixed

test/pool.test.ts
CodeSandwich: Fixed

test/pool.test.ts
CodeSandwich: Fixed

geigerzaehler (approved):

PR #45, opened 2020-10-15
Add nextCycle to receiver deltas

CodeSandwich:
Reduces gas usage when adding a receiver. It reduces edited slots per receiver from 4 to 2 when starting or stopping a payment. On the other hand it slightly increases gas usage while collecting, because more slots need to accessed.

When I have more time I need to dive deeper to understand the costs, there are many mysterious gas sinks, which I can't explain. E.g. reading a variable from a storage map costing 1000 gas. Or deletion of slots in a loop, which when commented out increases gas usage by just a few thousand instead of tens of thousands.

### DIFF
--- a/contracts/Pool.sol
+++ b/contracts/Pool.sol
@@ -37,6 +37,9 @@ pragma experimental ABIEncoderV2;
 /// only high-level abstractions for the user, Ethereum isn't really capable of scheduling work.
 /// The actual implementation emulates that behavior by calculating the results of the scheduled
 /// events based on how many blocks have been mined and only when a user needs their outcomes.
+///
+/// The contract assumes that all amounts in the system can be stored in signed 128-bit integers.
+/// It's guaranteed to be safe only when working with assets with supply lower than `2 ^ 127`.
 contract Pool {
     using ReceiverWeightsImpl for ReceiverWeights;
 
@@ -56,15 +59,15 @@ contract Pool {
         // Block number at which the funding period has started
         uint64 startBlock;
         // The amount available when the funding period has started
-        uint192 startBalance;
-        // --- SLOT BOUNDARY
-        // The target amount sent on each block.
-        // The actual amount is rounded down to the closes multiple of `weightSum`.
-        uint192 amtPerBlock;
+        uint128 startBalance;
         // The total weight of all the receivers
         uint32 weightSum;
         // The number of the receivers
         uint32 weightCount;
+        // --- SLOT BOUNDARY
+        // The target amount sent on each block.
+        // The actual amount is rounded down to the closes multiple of `weightSum`.
+        uint128 amtPerBlock;
         // --- SLOT BOUNDARY
         // The receivers' addresses and their weights
         ReceiverWeights receiverWeights;
@@ -73,12 +76,18 @@ contract Pool {
     struct Receiver {
         // The next block to be collected
         uint64 nextCollectedCycle;
-        // The amount of funds received for the last collected cycle
-        uint192 lastFundsPerCycle;
+        // The amount of funds received for the last collected cycle.
+        // It never is negative, it's a signed integer only for convenience of casting.
+        int128 lastFundsPerCycle;
         // --- SLOT BOUNDARY
         // The changes of collected amounts on specific cycle.
         // The keys are cycles, each cycle becomes collectable on block `C * cycleBlocks`
-        mapping(uint64 => int256) amtDeltas;
+        mapping(uint64 => AmtDelta) amtDeltas;
+    }
+
+    struct AmtDelta {
+        int128 thisCycle;
+        int128 nextCycle;
     }
 
     struct ReceiverWeight {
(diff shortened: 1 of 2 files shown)


contracts/Pool.sol
```diff
@@ -56,11 +59,11 @@ contract Pool {
         // Block number at which the funding period has started
         uint64 startBlock;
         // The amount available when the funding period has started
-        uint192 startBalance;
+        uint128 startBalance;
```
geigerzaehler, line 62: Does it make sense to reorder the fields to align the slot boundaries again?

contracts/Pool.sol
```diff
@@ -102,18 +110,20 @@ contract Pool {
...
     /// by the sender of the message
     /// @return collected The available amount
-    function collectable() public view returns (uint256) {
+    function collectable() public view returns (uint128) {
         Receiver storage receiver = receivers[msg.sender];
         uint64 collectedCycle = receiver.nextCollectedCycle;
         if (collectedCycle == 0) return 0;
+        uint256 currFinishedCycle = block.number / cycleBlocks;
+        if (collectedCycle > currFinishedCycle) return 0;
         uint256 collected = 0;
```
geigerzaehler, line 119: We don’t we declare this as uint128? We coerce it when we return anyways.

contracts/Pool.sol
```diff
@@ -127,21 +137,19 @@ contract Pool {
...
         int256 lastFundsPerCycle = receiver.lastFundsPerCycle;
         for (; collectedCycle <= currFinishedCycle; collectedCycle++) {
-            int256 delta = receiver.amtDeltas[collectedCycle];
-            if (delta != 0) {
-                delete receiver.amtDeltas[collectedCycle];
-                lastFundsPerCycle += delta;
-            }
```
geigerzaehler, line 134: I think it makes sense to keep this. If `delta == 0` and we delete the storage then this will cost gas. We only receive a refund from `delete` if the storage slot was non-zero.

contracts/Pool.sol
```diff
@@ -102,18 +110,20 @@ contract Pool {
...
     /// @return collected The available amount
-    function collectable() public view returns (uint256) {
+    function collectable() public view returns (uint128) {
         Receiver storage receiver = receivers[msg.sender];
         uint64 collectedCycle = receiver.nextCollectedCycle;
         if (collectedCycle == 0) return 0;
+        uint256 currFinishedCycle = block.number / cycleBlocks;
+        if (collectedCycle > currFinishedCycle) return 0;
         uint256 collected = 0;
         int256 lastFundsPerCycle = receiver.lastFundsPerCycle;
```
geigerzaehler, line 120: Probably makes sense to have this as uint128, too.

contracts/Pool.sol
CodeSandwich: I've been considering it, but have decided not to touch it. We can't reduce the number of used slots to 1. Alignment is maintained by Solidity, it packs the fields one after another, but never cuts any in half and zero-pads the necessary gaps. I like this order too, because it's possible to touch only one slot when topping up.

contracts/Pool.sol
CodeSandwich: This is something I'm hesitant about in general, we can either:

- A. Use the smallest possible possible integers fitting necessary data. This makes reasoning about code a bit easier, but requires Solidity to artificially maintain the lowered precision. All storage and operations on the EVM are performed on 256-bit values.
- B. Use 256-bit values everywhere possible. This obscures invariants about the possible values of variables, but aligns with the EVM architecture.

I honestly have no idea, which one is better. Now I'm thinking that maybe you're right and we should focus on core readability and let the optimizer take care of uglification and necessary casts?

contracts/Pool.sol
CodeSandwich: But now `delta` comes from both `receiver.amtDeltas[collectedCycle - 1].nextCycle` and `receiver.amtDeltas[collectedCycle].thisCycle;`. To decide whether or not to `delete receiver.amtDeltas[collectedCycle - 1];` we would need to either read `receiver.amtDeltas[collectedCycle - 1].thisCycle` too to confirm that it was non-zero, which costs 200 gas, exactly as much as setting storage from 0 to 0. I think that this is the cheapest option possible, any additional read would immediately obliterate any possible savings.

contracts/Pool.sol
CodeSandwich: But what we can do is cramming `weightSum` and `weightCount` into the first slot. This way only updates of `amtPerBlock` will need to touch the second slot.

CodeSandwich: Fixed

contracts/Pool.sol
geigerzaehler: This answers my question and I’d be ok with keeping it as it is but I’d still prefer to avoid unecessarry casts and use the type that we intend to use.

CodeSandwich: Fixed, reduced number of casts and slimmed down integer types to better match what they can actually contain

contracts/Pool.sol

geigerzaehler (approved):

Issue #19, opened 2020-09-28
Make pools ERC-20 compatible


CodeSandwich:
We need to be able to deploy a pool contract, which will be pushing donations in arbitrary ERC-20 tokens. The token contract can be passed during construction, it's fine to have deployed one per token.

geigerzaehler:
> We need to be able to deploy a pool contract, which will be pushing donations in arbitrary ERC-20 tokens.

This sounds like one pool contract can handle arbitrary ERC-20s. If I understand this correctly though we want the pool contract to be generic over the used ERC-20 token, right?

CodeSandwich:
Right, we want a contract to be generic over all ERC-20 tokens, but we want to attach it to exactly one token (ERC-20 contract address) on deployment.

PR #48, opened 2020-10-20
Add ERC-20 version of the pool

CodeSandwich:
Fixes https://github.com/radicle-dev/radicle-contracts/issues/19

### DIFF
--- a/contracts/Pool.sol
+++ b/contracts/Pool.sol
@@ -2,6 +2,8 @@
 pragma solidity ^0.6.2;
 pragma experimental ABIEncoderV2;
 
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
 /// @notice Funding pool contract. Automatically sends funds to a configurable set of receivers.
 ///
 /// The contract has 2 types of users: the senders and the receivers.
@@ -173,12 +175,18 @@ abstract contract Pool {
 
     /// @notice Withdraws unsent funds of the sender of the message and sends them to that sender
     /// @param amount The amount to be withdrawn, must not be higher than available funds
-    function withdraw(uint128 amount) public suspendPayments {
+    function withdraw(uint128 amount) public {
         if (amount == 0) return;
+        withdrawInternal(amount);
+        transferToSender(amount);
+    }
+
+    /// @notice Withdraws unsent funds of the sender of the message
+    /// @param amount The amount to be withdrawn, must not be higher than available funds
+    function withdrawInternal(uint128 amount) internal suspendPayments {
         uint128 startBalance = senders[msg.sender].startBalance;
         require(amount <= startBalance, "Not enough funds in the sender account");
         senders[msg.sender].startBalance = startBalance - amount;
-        transferToSender(amount);
     }
 
     /// @notice Sets the target amount sent on every block from the sender of the message.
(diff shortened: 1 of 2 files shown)


contracts/Pool.sol
```diff
@@ -391,6 +399,74 @@ contract EthPool is Pool {
...
+/// In order to send received tokens, they must be first `collect`ed and then `topUp`ped
+/// if they are to be sent through the contract.
+///
+/// The concept of something happening periodically, e.g. every block or every `cycleBlocks` are
+/// only high-level abstractions for the user, Ethereum isn't really capable of scheduling work.
+/// The actual implementation emulates that behavior by calculating the results of the scheduled
+/// events based on how many blocks have been mined and only when a user needs their outcomes.
+///
+/// The contract assumes that all amounts in the system can be stored in signed 128-bit integers.
+/// It's guaranteed to be safe only when working with tokens with supply lower than `2 ^ 127`.
```
cloudhead, line 440: This comment block is now in three places I believe. We should maybe move it to the top of the file, or have it on one contract and reference it from the others.

contracts/Pool.sol
```diff
@@ -391,6 +399,74 @@ contract EthPool is Pool {
...
+
+    /// @notice Tops up the sender balance of a sender of the message.
+    /// The contract will top up with all the tokens it has allowance for from the sender.
+    function topUp() public payable {
+        uint256 amount = erc20.allowance(msg.sender, address(this));
+        if (amount > 0) {
+            erc20.transferFrom(msg.sender, address(this), amount);
+            onTopUp(uint128(amount));
+        }
+    }
```
cloudhead, line 463: In practice, this function may be problematic to use, as it'll require 2 transactions to top up a certain amount (set allowance + topup), compared to only one for the Eth version. An alternative would be to pass in the amount, and in the Eth version, reimburse any difference between `msg.value` and the amount parameter. Users can then set a high allownace *once*, and then top-up whatever they want.

cloudhead: Another nit: since this function is payable, it's possible to lose money by sending ether to it. Perhaps it's a good idea to send back any ether to the sender in the erc20 version.

CodeSandwich: Oh, I forgot to remove `payable`, fixed. I'm not trying to make both versions of `TopUp` have the same API, that's why they aren't a part of the common `Pool`. IMO they're too different to unify.

Passing an `amount` in ERC-20 `TopUp` was my initial design, but I've decided to drop it. It required you to use the desired amount twice, once for allowance and once for `TopUp` and I really couldn't find a use case where these amounts would be different.

Why would somebody allow 10 tokens, but `TopUp` with only 6? Why would he want to wait for with the other 4? If he wants to `TopUp` with 4 next month, would preparing another allowance be so hard? I'm also quite wary with dangling allowances, ANYBODY can trigger it and effectively lock the tokens, because the pool can't notice random transactions.

contracts/Pool.sol
CodeSandwich: I agree, this is stupid. `@inheritdoc` would be perfect here, but Natspec doesn't support it for the contracts. AFAIK it doesn't have any means of referencing the other items either.

Fixed, now we're just telling the users to go and somehow find the docs for `Pool` by themselves, should be enough :shrug:

contracts/Pool.sol
CodeSandwich: Fixed

cloudhead (approved):
:+1:

PR #49, opened 2020-10-21
Encapsulate pool test utils

CodeSandwich:
The test tools have grown a lot recently in an organic way. Some of them require raw `address`es, some `PoolUser`s, there's also a lot of boilerplate around accessing `PoolUser`s' fields. This PR aims at cleaning it up by moving most of the tools from functions to `PoolUser`s' methods.

### DIFF
--- a/test/pool.test.ts
+++ b/test/pool.test.ts
@@ -23,61 +23,6 @@ import {
 
 const CYCLE_BLOCKS = 10;
 
-type AnyPool = EthPool | Erc20Pool;
-
-async function getEthPoolUsers(): Promise<EthPoolUser[]> {
-  const signers = await buidler.ethers.getSigners();
-  const pool = await new EthPoolFactory(signers[0]).deploy(CYCLE_BLOCKS);
-  await pool.deployed();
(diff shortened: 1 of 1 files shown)


test/pool.test.ts
```diff
@@ -23,379 +23,371 @@ import {
...
+async function mineBlocksUntilCycleEnd(): Promise<void> {
+  const blockNumber = await buidler.ethers.provider.getBlockNumber();
+  await mineBlocks(CYCLE_BLOCKS - ((blockNumber + 1) % CYCLE_BLOCKS));
+}
+
 type AnyPool = EthPool | Erc20Pool;
 
+abstract class PoolUser {
+  pool: AnyPool;
+  addr: string;
```
geigerzaehler, line 37: A comment to clarify that this is the address of the user (and not the pool) would be helpful.

geigerzaehler (approved):

test/pool.test.ts
CodeSandwich: Fixed

Issue #16, opened 2020-09-23
Add shared pool funding


CodeSandwich:
A user should be able to create and maintain a list of receivers, which others can reuse.

https://www.notion.so/monadic/Funding-Pools-protocol-35d73282538c4d21aa096753fea540e3

PR #52, opened 2020-11-02
Rework ReceiverWeights iterator

CodeSandwich:
The first step toward https://github.com/radicle-dev/radicle-contracts/issues/16. Alters the `ReceiverWeights` iterator by making its API dependant not on weight, but on receiver address. It's also tighter with memory usage and simpler when it comes to the `view` iterator.

### DIFF
--- a/contracts/Pool.sol
+++ b/contracts/Pool.sol
@@ -313,7 +313,7 @@ abstract contract Pool {
         while (true) {
             uint32 weight;
             (receiverAddr, weight) = sender.receiverWeights.nextWeightPruning(receiverAddr);
-            if (weight == 0) break;
+            if (receiverAddr == ReceiverWeightsImpl.ADDR_ROOT) break;
             Receiver storage receiver = receivers[receiverAddr];
             // The receiver was never used, initialize it
             if (amtPerWeightPerBlockDelta > 0 && receiver.nextCollectedCycle == 0)
@@ -418,8 +418,8 @@ library ReceiverWeightsImpl {
     /// Prunes all the zeroed items found between the current and the next receivers.
     /// Iterating over the whole list prunes all the zeroed items.
     /// @param current The previously returned receiver address or ADDR_ROOT to start iterating
-    /// @return next The next receiver address
-    /// @return weight The next receiver weight, 0 if the end of the list was reached
+    /// @return next The next receiver address, ADDR_ROOT if the end of the list was reached
+    /// @return weight The next receiver weight
     function nextWeightPruning(ReceiverWeights storage self, address current)
         internal
         returns (address next, uint32 weight)
@@ -435,7 +435,11 @@ library ReceiverWeightsImpl {
                     // Somehow it's ~1500 gas cheaper than `delete self[next]`
                     self.data[next].next = ADDR_UNINITIALIZED;
                     next = newNext;
-                    if (next == ADDR_END) break;
+                    if (next == ADDR_END) {
+                        // Removing the last item on the list, clear the storage
+                        if (current == ADDR_ROOT) next = ADDR_UNINITIALIZED;
+                        break;
+                    }
                     weight = self.data[next].weight;
                 } while (weight == 0);
                 // link the previous non-zero element with the next non-zero element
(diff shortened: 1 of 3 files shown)


contracts/Pool.sol
```diff
@@ -435,36 +435,34 @@ library ReceiverWeightsImpl {
...
+                        break;
+                    }
                     weight = self.data[next].weight;
                 } while (weight == 0);
                 // link the previous non-zero element with the next non-zero element
                 // or ADDR_END if it became the last element on the list
                 self.data[current].next = next;
             }
         }
+        if (next == ADDR_END) next = ADDR_ROOT;
```
geigerzaehler, line 450: Is there a reason we’re distinguishing between `ADDR_END` and `ADDR_ROOT`? This pattern here is common with `nextWeight` and `setWeight`. It may be possible that we don’t need to distinguish these.

test/pool.test.ts
```diff
@@ -738,6 +738,30 @@ describe("ReceiverWeights", function () {
...
+    // Add an item and then clear the list
+    const weights_test = await deployReceiverWeightsTest();
+    const [addr1] = randomAddresses();
+
+    await weights_test.setWeights([
+      {receiver: addr1, weight: 1},
+      {receiver: addr1, weight: 0},
+    ]);
+
+    assert((await weights_test.receiverWeightsSumDelta()).eq(1 - 1));
```
geigerzaehler, line 751:
```suggestion
    assert((await weights_test.receiverWeightsSumDelta()).eq(0));
```

test/pool.test.ts
```diff
@@ -738,6 +738,30 @@ describe("ReceiverWeights", function () {
...
+    assert(weights.length == 0);
+
+    // Add an item
+    await weights_test.setWeights([{receiver: addr1, weight: 2}]);
+
+    assert((await weights_test.receiverWeightsSumDelta()).eq(2));
+    weights = await weights_test.getReceiverWeightsIterated();
+    assert(weights.length == 1);
+    assert(weights[0].receiver == addr1);
+    assert(weights[0].weight == 2);
```
geigerzaehler, line 762:
```suggestion
    assert.deepEqual(weights, { receiver: addr1, weight: 2 });
```

contracts/Pool.sol
CodeSandwich: Unfortunately we need both of them. They differentiate between a slot for a weight, which may be empty, but is attached to the linked list (`ADDR_END`) and a slot, which is empty and not attached (`ADDR_ROOT` aka `ADDR_UNINITIALIZED`). It prevents the list from creating a cycle when the last item is zeroed and then assigned a value.

test/pool.test.ts
CodeSandwich: This explains, where the value comes from. Here it's trivial, but it follows a pattern, which in other tests prevents unexpected values from popping up. I think that it makes no harm here, but it preserves the convention.

test/pool.test.ts
CodeSandwich: I couldn't use it in code, could you help me out with this?

`assert.deepEqual(weights, [{receiver: addr1, weight: 2}]);`
causes a compilation error:
`Type '{ receiver: string; weight: number; }' is missing the following properties from type '{ receiver: string; weight: number; 0: string; 1: number; }': 0, 1`

So I guess that the silly fix for that should be:
`assert.deepEqual(weights, [{receiver: addr1, weight: 2, 0: addr1, 1: 2}]);`
it compiles, but now the test fails:
```
      AssertionError: expected [ Array(1) ] to deeply equal [ Array(1) ]
      + expected - actual

       [
      -  [
      -    "0x1B78D18678cfA961f98E7DC824e9ABD046b0c594"
      -    2
      -  ]
      +  {
      +    "0": "0x1B78D18678cfA961f98E7DC824e9ABD046b0c594"
      +    "1": 2
      +    "receiver": "0x1B78D18678cfA961f98E7DC824e9ABD046b0c594"
      +    "weight": 2
      +  }
       ]
```

Ok, an array then:
`assert.deepEqual(weights, [[addr1, 2]]);`
but it again fails to compile with:
`Type '[string, number]' is not assignable to type '{ receiver: string; weight: number; 0: string; 1: number; }'.`

geigerzaehler: I see. I assumed etherjs just returned an object. But if it returns this more complicated struct I’m ok if we leave it like it is.

contracts/Pool.sol
geigerzaehler: I think I’ve finally understood it. The distinction only becomes important in `setWeights` where we need to decide whether to insert or to update in place.

geigerzaehler (approved):

PR #54, opened 2020-11-05
Add ProxyDeltas data structure

CodeSandwich:
Adds `ProxyDeltas`, which is a random access, iterable mapping between cycles and deltas of funds received by a proxy. It will be used when updating a proxy to determine how to update the downstream receivers' delta. It's an iterable version of a receiver's list of deltas, it stores the same data. Implementation and tests are a slightly modified version of the `ReceiverWeights` linked list. Some form of generics would be great in Solidity :shrug: `ProxyDeltas` aren't used in the production code yet, the next PR will do that, this PR is already quite big.

### DIFF
--- a/contracts/Pool.sol
+++ b/contracts/Pool.sol
@@ -3,6 +3,8 @@ pragma solidity ^0.6.2;
 pragma experimental ABIEncoderV2;
 
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "./libraries/ProxyDeltas.sol";
+import "./libraries/ReceiverWeights.sol";
 
 /// @notice Funding pool contract. Automatically sends funds to a configurable set of receivers.
 ///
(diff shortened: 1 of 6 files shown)


contracts/Pool.sol
```diff
@@ -529,3 +529,105 @@ library ReceiverWeightsImpl {
...
     }
 }
+
+/// @notice A list of cycles and their deltas of amounts received by a proxy.
+/// For each cycle there are stored two deltas, one for the cycle itself
+/// and one for the cycle right after it.
+/// It reduces storage access for some usage scenarios.
+/// The cycle is described by its own entry and an entry for the previous cycle.
+/// Iterable and with random access.
+struct ProxyDeltas {
```
geigerzaehler, line 539: Does it make sense to split this out into a separate file?

contracts/Pool.sol
```diff
@@ -529,3 +529,105 @@ library ReceiverWeightsImpl {
...
+    uint64 internal constant CYCLE_UNINITIALIZED = 0;
+    uint64 internal constant CYCLE_END = type(uint64).max;
+
+    /// @notice Return the next non-zero, non-obsolete delta and its cycle.
+    /// The order is undefined, it may or may not be chronological.
+    /// Prunes all the fully zeroed or obsolete items found between the current and the next cycle.
+    /// Iterating over the whole list prunes all the zeroed and obsolete items.
+    /// @param current The previously returned cycle or CYCLE_ROOT to start iterating
+    /// @param finishedCycle The last finished cycle.
+    /// Data describing only the finished cycles is considered obsolete.
```
geigerzaehler, line 563: Suggestion: I think the following would make the docs a bit clearer

```suggestion
    /// Entries for cycles before `finishedCycle` are considered obsolete.
```

contracts/Pool.sol
```diff
@@ -529,3 +529,105 @@ library ReceiverWeightsImpl {
...
+/// The list uses 2 words of storage per stored cycle.
+library ProxyDeltasImpl {
+    struct ProxyDeltaStored {
+        uint64 next;
+        // --- SLOT BOUNDARY
+        Pool.AmtDelta delta;
+    }
+
+    uint64 internal constant CYCLE_ROOT = 0;
+    uint64 internal constant CYCLE_UNINITIALIZED = 0;
```
geigerzaehler, line 554: For me, the code would become more readable if we would use only one name for the same constant value. Otherwise one might be lead to believe that this are different values.

But this could also be a follow-up where we make the same change to to `ReceiverWeights`.

test/proxy.deltas.test.ts
```diff
@@ -0,0 +1,372 @@
...
+import buidler from "@nomiclabs/buidler";
+import {BigNumber, BigNumberish} from "ethers";
+import {assert} from "chai";
+import {submitFailing} from "./support";
+
+async function deployProxyDeltasTest(): Promise<ProxyDeltasTest> {
+  const [signer] = await buidler.ethers.getSigners();
+  return new ProxyDeltasTestFactory(signer)
+    .deploy()
+    .then((deltas) => deltas.deployed());
```
geigerzaehler, line 12: I’d prefer to use `await` consistently.

```suggestion
  const deltas = await new ProxyDeltasTestFactory(signer).deploy();
  await deltas.deployed();
```

test/proxy.deltas.test.ts
```diff
@@ -0,0 +1,372 @@
+import {ProxyDeltasTestFactory} from "../contract-bindings/ethers";
```
geigerzaehler, line 1: Is there a reason why this file is named `proxy.deltas.test.ts` instead of `proxy-deltas.test.ts`?

test/proxy.deltas.test.ts
```diff
@@ -0,0 +1,372 @@
...
+import {submitFailing} from "./support";
+
+async function deployProxyDeltasTest(): Promise<ProxyDeltasTest> {
+  const [signer] = await buidler.ethers.getSigners();
+  return new ProxyDeltasTestFactory(signer)
+    .deploy()
+    .then((deltas) => deltas.deployed());
+}
+
+async function expectAddToDeltasWithInvalidCycleReverts(
```
geigerzaehler, line 15: Personally, I’d inline this function to avoid unecessary indirection.

geigerzaehler (commented):
Again, great work :construction_worker:

test/proxy.deltas.test.ts
CodeSandwich: It's used in 2 places. I've decided to make it DRY, because it contains some non-obvious code and hardcoded copies of strings from the contract. I don't feel very strong about keeping it like this.

test/proxy.deltas.test.ts
CodeSandwich: None at all, fixed

test/proxy.deltas.test.ts
CodeSandwich: Me too, fixed

contracts/Pool.sol
CodeSandwich: I kind of agree. I don't like the fact that the cycle number `0` has 2 different meanings depending on context, but you're right that having 2 names for the same value is madness. And no, they can't have different values, `CYCLE_ROOT` must equal `CYCLE_UNINITIALIZED`, which must equal `0`. Fixed.

contracts/Pool.sol
CodeSandwich: Fixed.

geigerzaehler (approved):
:arrow_forward:

PR #57, opened 2020-11-11
Pool proxy

CodeSandwich:
Fixes https://github.com/radicle-dev/radicle-contracts/issues/16. It's pretty big and complex, but it's not dividable further in any reasonable way, sorry.

### DIFF
--- a/contracts/Pool.sol
+++ b/contracts/Pool.sol
@@ -8,30 +8,38 @@ import "./libraries/ReceiverWeights.sol";
 
 /// @notice Funding pool contract. Automatically sends funds to a configurable set of receivers.
 ///
-/// The contract has 2 types of users: the senders and the receivers.
+/// The contract has 3 types of users: the senders, the receivers and the proxies.
 ///
-/// A sender has some funds and a set of addresses of receivers, to whom he wants to send funds.
+/// A sender has some funds and a set of addresses of receivers, to whom he wants to send funds
+/// and proxies, who he trusts with choice of more receivers.
 /// In order to send there are 3 conditions, which must be fulfilled:
 ///
 /// 1. There must be funds on his account in this contract.
 ///    They can be added with `topUp` and removed with `withdraw`.
 /// 2. Total amount sent to the receivers on each block must be set to a non-zero value.
 ///    This is done with `setAmountPerBlock`.
-/// 3. A set of receivers must be non-empty.
-///    Receivers can be added, removed and updated with `setReceiver`.
-///    Each receiver has a weight, which is used to calculate how the total sent amount is split.
+/// 3. A set of receivers and proxies must be non-empty.
+///    Receivers can be added, removed and updated with `setReceiver`, proxies with `setProxy`.
+///    Each receiver or proxy has a weight,
+///    which is used to calculate how the total sent amount is split.
 ///
 /// Each of these functions can be called in any order and at any time, they have immediate effects.
 /// When all of these conditions are fulfilled, on each block the configured amount is being sent.
 /// It's extracted from the `withdraw`able balance and transferred to the receivers.
 /// The process continues automatically until the sender's balance is empty.
 ///
-/// The receiver has an account, from which he can `collect` funds sent by the senders.
+/// A receiver has an account, from which he can `collect` funds sent by the senders.
 /// The available amount is updated every `cycleBlocks` blocks,
 /// so recently sent funds may not be `collect`able immediately.
 /// `cycleBlocks` is a constant configured when the pool is deployed.
 ///
-/// A single address can be used both as a sender and as a receiver.
+/// A proxy has a list of receivers with attached weights, just like a sender.
+/// When a sender is sending some funds to a proxy,
+/// they are split and passed to the proxy's receivers according to their weights.
+/// The proxy can change its list of receivers at any time with immediate effect,
+/// but as if it was done on the beginning of the current cycle.
+///
+/// A single address can be used as a sender, a receiver and a proxy, even at the same time.
 /// It will have 2 balances in the contract, one with funds being sent and one with received,
 /// but with no connection between them and no shared configuration.
 /// In order to send received funds, they must be first `collect`ed and then `topUp`ped
@@ -46,6 +54,7 @@ import "./libraries/ReceiverWeights.sol";
 /// It's guaranteed to be safe only when working with assets with supply lower than `2 ^ 127`.
 abstract contract Pool {
     using ReceiverWeightsImpl for ReceiverWeights;
+    using ProxyDeltasImpl for ProxyDeltas;
 
     /// @notice On every block `B`, which is a multiple of `cycleBlocks`, the receivers
     /// gain access to funds collected on all blocks from `B - cycleBlocks` to `B - 1`.
@@ -54,10 +63,19 @@ abstract contract Pool {
     uint64 internal constant MAX_BLOCK_NUMBER = type(uint64).max - 2;
     /// @notice Maximum sum of all receiver weights of a single sender.
     /// Limits loss of per-block funding accuracy, they are always multiples of weights sum.
-    uint32 public constant SENDER_WEIGHTS_SUM_MAX = 1000;
+    uint32 public constant SENDER_WEIGHTS_SUM_MAX = 10000;
     /// @notice Maximum number of receivers of a single sender.
     /// Limits costs of changes in sender's configuration.
     uint32 public constant SENDER_WEIGHTS_COUNT_MAX = 100;
+    /// @notice The sum of all receiver weights of a single proxy.
+    /// It must remain constant throughout the whole life of a proxy.
+    /// A sender's weight of a proxy must always be a multiple of this value.
+    /// Limits loss of per-block funding accuracy.
+    uint32 public constant PROXY_WEIGHTS_SUM = 100;
+    /// @notice Maximum number of receivers of a single proxy.
+    /// Whenever a proxy is added to a sender, this number is added to its `weightCount`.
+    /// Limits costs of changes in sender's or proxy's configuration.
+    uint32 public constant PROXY_WEIGHTS_COUNT_MAX = 10;
     /// @notice The amount passed to `withdraw` to withdraw all the funds
     uint128 public constant WITHDRAW_ALL = type(uint128).max;
 
@@ -66,9 +84,12 @@ abstract contract Pool {
         uint64 startBlock;
         // The amount available when the funding period has started
         uint128 startBalance;
-        // The total weight of all the receivers
+        // The total weight of all the receivers, must never be larger than `SENDER_WEIGHTS_SUM_MAX`
         uint32 weightSum;
-        // The number of the receivers
+        // The number of the receivers, must never be larger than `SENDER_WEIGHTS_COUNT_MAX`.
+        // A single address may serve as both a receiver and a proxy.
+        // Each proxy is counted as `PROXY_WEIGHTS_COUNT_MAX`,
+        // because that's how many actual receivers it may represent.
         uint32 weightCount;
         // --- SLOT BOUNDARY
         // The target amount sent on each block.
@@ -91,6 +112,14 @@ abstract contract Pool {
         mapping(uint64 => AmtDelta) amtDeltas;
     }
 
+    struct Proxy {
+        // The receivers' addresses and their weights
+        ReceiverWeights receiverWeights;
+        // --- SLOT BOUNDARY
+        // The changes of amount per weight received from all senders on a specific cycle.
+        ProxyDeltas amtPerWeightDeltas;
+    }
+
     struct AmtDelta {
         // Amount delta applied on this cycle
         int128 thisCycle;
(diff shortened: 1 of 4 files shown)


contracts/Pool.sol
```diff
@@ -349,6 +547,73 @@ abstract contract Pool {
...
+        if (receiversCount == 0) return;
+
+        // Iterating over deltas, see `ProxyDeltas` for details
+        uint64 cycle = ProxyDeltasImpl.CYCLE_ROOT;
+        uint64 finishedCycle = blockNumber / cycleBlocks;
+        uint64 currCycle = finishedCycle + 1;
+        int128 totalDelta = 0;
+        while (true) {
+            int128 thisCycleDelta;
+            int128 nextCycleDelta;
```
geigerzaehler, line 590: How do you feel about using `cycleDelta` and `followingCycleDelta` to avoid confusion with the `currCycle`? The “this” prefix seems to to indicate that we’re talking about the current cycle.

contracts/Pool.sol
```diff
@@ -349,6 +547,73 @@ abstract contract Pool {
...
+            (cycle, thisCycleDelta, nextCycleDelta) = proxy.amtPerWeightDeltas.nextDeltaPruning(
+                cycle,
+                finishedCycle
+            );
+            if (cycle > finishedCycle) thisCycleDelta *= multiplier;
+            else thisCycleDelta = 0;
+            nextCycleDelta *= multiplier;
+            if (cycle == ProxyDeltasImpl.CYCLE_ROOT) break;
+            totalDelta += nextCycleDelta;
+            if (cycle > finishedCycle) totalDelta += thisCycleDelta;
```
geigerzaehler, line 600: I’ll leave it up to you, but personally I find conditionals easier to read if there using braces and the statements are on the next line. Especially for `if/else`.

contracts/Pool.sol
```diff
@@ -349,6 +547,73 @@ abstract contract Pool {
...
+            int128 thisCycleDelta;
+            int128 nextCycleDelta;
+            (cycle, thisCycleDelta, nextCycleDelta) = proxy.amtPerWeightDeltas.nextDeltaPruning(
+                cycle,
+                finishedCycle
+            );
+            if (cycle > finishedCycle) thisCycleDelta *= multiplier;
+            else thisCycleDelta = 0;
+            nextCycleDelta *= multiplier;
+            if (cycle == ProxyDeltasImpl.CYCLE_ROOT) break;
```
geigerzaehler, line 598: Can we re-arrange the `if` statements to make them easier to follow? The `if/break` statement here should come as early as possible. Then it should also be possible to merge the two remaining `if` statements.

contracts/Pool.sol
```diff
@@ -349,6 +547,73 @@ abstract contract Pool {
...
+        uint64 currCycle = finishedCycle + 1;
+        int128 totalDelta = 0;
+        while (true) {
+            int128 thisCycleDelta;
+            int128 nextCycleDelta;
+            (cycle, thisCycleDelta, nextCycleDelta) = proxy.amtPerWeightDeltas.nextDeltaPruning(
+                cycle,
+                finishedCycle
+            );
+            if (cycle > finishedCycle) thisCycleDelta *= multiplier;
```
geigerzaehler, line 595: Doesn’t `nextDeltaPruning` ensure that `cycle > finishedCycle` because it prunes and skips obsolete cycles?

CodeSandwich: It ensures that `cycle >= finishedCycle`, we still need `nextCycleDelta` when `cycle == finishedCycle`

contracts/Pool.sol
CodeSandwich: Yeah, I took a look at this part and it was a completely uncleaned development mess, fixed :grimacing:

contracts/Pool.sol
CodeSandwich: I agree for `if/else`, cleaned up. But overall I think that oneliner `if`s are usually less distracting than full blown bracketed blocks. There's nothing in between, the formatter doesn't allow unbracketed two line `if`s.

contracts/Pool.sol
CodeSandwich: I think that overall having `this` and `next` is a clean and concise convention. It highlights that there are two deltas relative to a containing data structure key and to each other, but independent from the current block. I took a look in the code and general removal of `this` convention would require some wrestling with renaming of suddenly vague and meaningless variable names.

contracts/Pool.sol
```diff
@@ -307,30 +435,100 @@ abstract contract Pool {
...
+            if (receiver == ReceiverWeightsImpl.ADDR_ROOT) break;
+            setReceiverDeltaFromNow(receiver, perBlockPerProxyWeightDelta * weight, blockEnd);
+        }
+    }
+
+    /// @notice Sets delta of a single proxy on a given block number
+    /// @param proxyDeltas The deltas of the per-cycle receiving rate
+    /// @param blockNumber The block number from which the delta takes effect
+    /// @param perBlockDelta Change of the per-block receiving rate
+    function setSingleProxyDelta(
```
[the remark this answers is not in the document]

contracts/Pool.sol
```diff
@@ -307,30 +435,100 @@ abstract contract Pool {
...
+    /// @param blockEnd The block number from which the delta stops taking effect
+    function setReceiverDeltaFromNow(
+        address receiverAddr,
+        int128 perBlockDelta,
+        uint64 blockEnd
+    ) internal {
+        uint64 blockNumber = uint64(block.number);
+        Receiver storage receiver = receivers[receiverAddr];
+        // The receiver was never used, initialize it
+        if (perBlockDelta > 0 && receiver.nextCollectedCycle == 0)
```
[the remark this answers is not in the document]
CodeSandwich: Definitely not, `perBlockDelta` is always negative when one is stopping sending in `suspendProxy`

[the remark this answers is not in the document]
geigerzaehler: Maybe it’s worth adding a comment why this is not just `receiver.NextCollectedCycle == 0`: `perBlockDelta <= 0` can only happen if the receiver has been initialized.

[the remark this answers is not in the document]
CodeSandwich: Good point, commented

test/pool.test.ts
```diff
@@ -28,6 +28,41 @@ async function mineBlocksUntilCycleEnd(): Promise<void> {
...
+  }
+
+  isEmpty(): boolean {
+    return this.receiverWeight == 0 && this.proxyWeight == 0;
+  }
+}
+
+type ProxyReceiverWeights = Map<string, ProxyReceiverWeight>;
+
+function updateReceiversList(
```
geigerzaehler, line 51:
```suggestion
function updateReceiverWeights(
```
This would align the noun in the function name with the type of the subject and make it clearer what is going on. It also confused me that we were using the term “list” whereas the actual type is a map.

test/pool.test.ts
```diff
@@ -28,6 +28,41 @@ async function mineBlocksUntilCycleEnd(): Promise<void> {
...
+  list: ProxyReceiverWeights,
+  receiver: PoolUser,
```
geigerzaehler, line 53: We don’t need any information from `PoolUser` except for the address.
```suggestion
  receiver: string,
```
I’d prefer this approach because the function signature looks much more like your usual update function with object-key-function arguments.

test/pool.test.ts
```diff
@@ -45,30 +80,103 @@ abstract class PoolUser {
...
+    weight: number,
+    expectedCause: string
+  ): Promise<void> {
+    const receivers = await this.getAllReceivers();
+    await submitFailing(
+      this.pool.setProxy(receiver.addr, weight),
+      "setProxy",
+      expectedCause
+    );
+    await this.expectReceivers(receivers);
```
geigerzaehler, line 130: I don’t think this check is necessary. This is guaranteed by the EVM and slows down the test.

test/pool.test.ts
```diff
@@ -45,30 +80,103 @@ abstract class PoolUser {
...
+    const receivers = await this.pool.getAllReceivers();
+    return new Map(
+      receivers.map(({receiver, receiverWeight, proxyWeight}) => [
+        receiver,
+        new ProxyReceiverWeight(receiverWeight, proxyWeight),
+      ])
+    );
+  }
+
+  async setProxyWeights(weights: [this, number][]): Promise<void> {
```
geigerzaehler, line 143: Similar to what I said above: We only need the address from each weight. To avoid confusion I’d prefer
```suggestion
  async setProxyWeights(weights: [string, number][]): Promise<void> {
```

test/pool.test.ts
```diff
@@ -522,6 +648,393 @@ describe("EthPool", function () {
     }
     await sender.expectSetReceiverReverts(receiver, 1, "Too many receivers");
   });
+
+  it("Allows sending via a proxy", async function () {
+    const [sender, proxy, receiver] = await getEthPoolUsers();
+    const proxyWeightBase = await sender.pool.PROXY_WEIGHTS_SUM();
+    await proxy.setProxyWeights([[receiver, proxyWeightBase]]);
+    await sender.setProxy(proxy, proxyWeightBase);
+    await sender.setAmountPerBlock(proxyWeightBase);
```
geigerzaehler, line 657: Can we choose a different multiple of `proxyWeightBase` here? This would make the relationship between the value and the proxy weight base clearer.

contracts/Pool.sol
```diff
@@ -347,6 +549,72 @@ abstract contract Pool {
...
+            receiversList[receiversCount++] = ReceiverWeight(receiverAddr, weight);
+        }
+        // The proxy doesn't exist
+        if (receiversCount == 0) return;
+
+        // Iterating over deltas, see `ProxyDeltas` for details
+        uint64 cycle = ProxyDeltasImpl.CYCLE_ROOT;
+        uint64 finishedCycle = blockNumber / cycleBlocks;
+        uint64 currCycle = finishedCycle + 1;
+        int128 totalDelta = 0;
```
geigerzaehler, line 589: After spending some time with this I think the following comment would clear things up.

```suggestion
        // The sum off all future changes to the per-block amount the proxy receives. If
        // `multiplier` is 1, then this is necessarily the per-block amount the proxy receives in
        // the current cycle—otherwise it’s the inverse.
        int128 totalDelta = 0;
```

geigerzaehler (commented):
I want to give this a final pass this afternoon

contracts/libraries/ReceiverWeights.sol
```diff
@@ -97,6 +97,14 @@ library ReceiverWeightsImpl {
         if (next == ADDR_END) next = ADDR_ROOT;
     }
 
+    /// @notice Checks if the list is fully zeroed and takes no storage space.
+    /// It means that either it was never used or that
+    /// it's been pruned after removal of all the elements.
+    /// @return True if the list is zeroed
+    function isZeroed(ReceiverWeights storage self) internal view returns (bool) {
```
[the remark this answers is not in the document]
CodeSandwich: It's a little different and I wanted to avoid confusion. The list can be empty and its iterator will yield nothing, but still hold some data. The real `isEmpty` would need to use the iterator, but that would be more expensive than this simplified `isZeroed`, which is enough for this use case.

contracts/Pool.sol
CodeSandwich: Cool, thank you! I've redacted it slightly:
```
        // The sum of all the future changes to the per-block amount the proxy receives.
        // This is also the per-block amount the proxy receives per weight in the current cycle,
        // but with its sign inverted.
        // Thus if `multiplier` is `1`, then this value is negative and if `-1`, it's positive.
```

test/pool.test.ts
CodeSandwich: Renamed.

This might be a mistake, but this data structure is called a `list` consistently throughout the codebase. The underlying data structure is a hybrid between a map and a linked list. I agree that `map` would make slightly more sense, because its API resembles a classic map with its free key-based access. If that's confusing, I can rename it, but in a separate PR. What do you think?

test/pool.test.ts
CodeSandwich: I'd very much prefer not to do that. This will require extracting the address for every proxy receiver in every test, which is a ton of unnecessary noise in the test bodies.

test/pool.test.ts
CodeSandwich: Good catch :D Fixed

test/pool.test.ts
CodeSandwich: Fixed

test/pool.test.ts
CodeSandwich: Fixed

test/pool.test.ts
geigerzaehler: Let’s leave it like it is. There’s no clear case for either “list” or “map”.

test/pool.test.ts
geigerzaehler: Fair enough

geigerzaehler (approved):
Great work @CodeSandwich! :sparkles:

Issue #55, opened 2020-11-06
api: Make withdraw argument optional


NunoAlexandre:
At the moment, the `withdraw` function takes an argument that specifies how much to withdraw.

As discussed by call, having the user specifying that amount is dangerous since the user might specify, say, the "current" pool remaining balance (i.e, the user wants to withdraw all the funds), but it might occur that by the time the withdraw transaction is processed that said balance has reduced, causing the transaction to fail. That'd mean the user would pay for that transaction and get nothing out of it.

My suggestion is to make that argument optional. If not specified, all the remaining funds available by the time the tx is processed are withdrawn. In other words, by default `withdraw`ing would mean "withdraw everything".

The idea with making it optional is to still allow the user to specify a custom amount. We could offer only the "withdraw everything" variant in the first release and "withdraw x" in an upcoming release.

@CodeSandwich thoughts?

PR #58, opened 2020-11-17
Allow withrawal of all funds

CodeSandwich:
Fixes https://github.com/radicle-dev/radicle-contracts/issues/55.

The `WITHDRAW_ALL` value is not `0`, because in the upcoming batching API we'll need to somehow convey that the user doesn't want to withdraw anything. Solidity doesn't make it easy to have optional arguments in the APIs, so the value `0` looks like the best bet to express that.

### DIFF
--- a/test/pool.test.ts
+++ b/test/pool.test.ts
@@ -522,6 +522,31 @@ describe("EthPool", function () {
     }
     await sender.expectSetReceiverReverts(receiver, 1, "Too many receivers");
   });
+
+  it("Allows withdrawal of all funds", async function () {
+    const [sender, receiver] = await getEthPoolUsers();
+    const withdrawAll = await sender.pool.WITHDRAW_ALL();
+    await sender.topUp(0, 10);
+    await sender.setAmountPerBlock(1);
+    await sender.setReceiver(receiver, 1);
+    await mineBlocks(4);
+    const balanceBefore = await sender.pool.signer.getBalance();
+    await submit(
+      sender.pool.withdraw(withdrawAll, {gasPrice: 0}),
+      "withdraw ETH"
+    );
+    const balanceAfter = await sender.pool.signer.getBalance();
+    const withdrawn = balanceAfter.sub(balanceBefore).toNumber();
+    // Sender had 5 blocks paying 1 per block
+    expect(withdrawn).to.equal(
+      5,
+      "The withdrawn amount is different from the requested amount"
+    );
+    await sender.expectWithdrawable(0);
+    await mineBlocksUntilCycleEnd();
+    // Receiver had 5 blocks paying 1 per block
+    await receiver.collect(5);
+  });
 });
 
 describe("Erc20Pool", function () {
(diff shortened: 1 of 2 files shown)


NunoAlexandre (commented):
Thanks, Igor! :smiley: 

 Now, if we want to withdraw all the collectable funds, we'd need to first import `WITHDRAW_ALL` and pass that to mean "collect all funds".

I'd prefer to have a cleaner API as a client. Either:

1. Offer both `withdraw(uint128)` and `withdrawAll()`, with the latter calling the first with `WITHRAW_ALL` as an argument, freeing the client from that responsibility and making the API more ergonomic

2. Drop the `uint128` argument and just call `withdrawInternal` with `WITHDRAW_ALL`. We don't foresee any use case to support a partial withdraw so this solution is cleanest and checks all boxes.

test/pool.test.ts
```diff
@@ -522,6 +522,25 @@ describe("EthPool", function () {
...
   });
+
+  it("Allows withdrawal of all funds", async function () {
+    const [sender] = await getEthPoolUsers();
+    const withdrawAll = await sender.pool.WITHDRAW_ALL();
+    const amount = 1234;
+    await sender.topUp(0, amount);
+    const balanceBefore = await sender.pool.signer.getBalance();
+    await submit(
+      sender.pool.withdraw(withdrawAll, {gasPrice: 0}),
```
geigerzaehler, line 533: I’d feel more comfortable with the test if we also included some receivers to cover a reduction in the balance. However, if this is ok for you we can also leave as it is.

geigerzaehler (approved):

CodeSandwich:
> I'd prefer to have a cleaner API as a client.

I've been considering 1, but it's not going to go well with batching. If we can withdraw some or withdraw all or withdraw nothing, we'll have to start juggling with enums and optional parameters. We need an API, which can transform straight to the batching one.

I'm not convinced that we want to drop partial withdrawals, we were discussing this during the last sync and to me it seemed like we all agreed that this elasticity is important for the users. If we drop it, we'll force users to double time and cost if they want to withdraw, but not everything.

NunoAlexandre:
> > I'd prefer to have a cleaner API as a client.
> 
> I've been considering 1, but it's not going to go well with batching. If we can withdraw some or withdraw all or withdraw nothing, we'll have to start juggling with enums and optional (ignored) parameters. We need an API, which can transform straight to the batching one.
> 
> I'm not convinced that we want to drop partial withdrawals, we were discussing this during the last sync and to me it seemed like we all agreed that this elasticity is important for the users. If we drop it, we'll force users to double time and cost if they want to withdraw, but not everything.

Can you explain how this withdrawing feature touches the batching feature? The latter should only involve setting the receivers, the monthly contribution, and the initial balance (top up). Am I missing something? :thinking:

CodeSandwich:
Sorry, batching and batching :D I mean the API which allows you to set up or update the whole sender in one shot.

NunoAlexandre (approved):
Let's go as-is :surfer:

test/pool.test.ts
CodeSandwich: Fixed

NunoAlexandre:
> I'm not convinced that we want to drop partial withdrawals, we were discussing this during the last sync and to me it seemed like we all agreed that this elasticity is important for the users. If we drop it, we'll force users to double time and cost if they want to withdraw, but not everything.

For the record: sorry, I was confusing this with the collect funds feature :see_no_evil: good that you didn't.

PR #60, opened 2020-11-24
Add how pool works docs

CodeSandwich:
The hight level description of how the pool works. Please roast until easily digestible.

[Rendered](https://github.com/radicle-dev/radicle-contracts/blob/how_pool_works/docs/how_pool_works.md)


docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
+This document is an introduction to understanding how the funding pool contract works.
+It doesn't describe all the implementation details or the API,
+it focuses on the basic principles of the core mechanics.
+
+# Overview
+
+The funding pool is a smart contract, which creates real-time streams of donations.
+One can start, alter or end the process of sending his funds at any time with immediate effect.
```
lftherios, line 8: "of sending their funds"

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+The flow of funds is automatically maintained and steady over time.
+
+There are 3 roles present in the contract.
+Any Ethereum address can be used in any number of these roles at the same time.
+
+- The sender - has assets and chooses who does he want to send them to, how much and how fast
+- The receiver - receives funds from senders
+- The proxy - receives funds from senders, but immediately passes them to receivers of his choice
```
lftherios, line 16: "of their choice"

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+On every block, a sender can send funds to a receiver.
+The receiver can collect funds sent on a given block only when the cycle containing it is over.
+
+![](how_pool_works_2.png)
+
+Here we have a history of a receiver, who has 2 senders.
+Each of them was sending different amounts over different periods of time.
+On the end of each cycle, the collectable amount was increased by the total sent amount.
+
+Each receiver needs to know, how much was sent to him.
```
lftherios, line 42: "was sent to them"

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+To accomplish that the history of received funds needs to be stored in some form.
+The receiver doesn't care, who exactly is sending on each block and in what configuration.
+The only thing that matters are the collectable per-cycle values below the timeline.
+This reduces the amount of stored data significantly, but this can be optimized further.
+
+![](how_pool_works_3.png)
+
+In this example, we started with having the raw collectable value of 23 for every block until
+the end of the blockchain.
```
lftherios, line 51: what do you mean here "until the end of the blockchain?"

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+Now we need to describe only cycles when receiving anything.
+The senders usually are sending constant per-cycle amounts over long periods of time, so
+the added values tend to create long series of constant numbers, in this case, 5s.
+We exploit that and turn them into deltas relative to the previous cycles.
+Now we need deltas only for cycles where the sending rate changes, it's very cheap.
+This is what the contract actually stores: a mapping from cycle numbers to deltas.
+
+## Collecting
+
+When the receiver wants to collect the funds, he can iterate over the finished cycles,
```
lftherios, line 62: "they can iterate over"

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+reconstruct the added values from deltas and then calculate the final currently collectable value.
```
lftherios, line 63: maybe better "the collectable amount at this point in time"

CodeSandwich: Fixed

docs/how_pool_works.md
CodeSandwich: Fixed

docs/how_pool_works.md
CodeSandwich: I mean for infinite number of blocks higher than this one, like in "until the end of time". The blockchain obviously has no end :D

docs/how_pool_works.md
CodeSandwich: Fixed

docs/how_pool_works.md
CodeSandwich: Fixed

docs/how_pool_works.md
CodeSandwich: Fixed

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+
+In this example, we started with having the raw collectable value of 23 for every block until
+the end of the blockchain.
+We then reduced that to storing values added to the collectable amount on each cycle.
+Now we need to describe only cycles when receiving anything.
+The senders usually are sending constant per-cycle amounts over long periods of time, so
+the added values tend to create long series of constant numbers, in this case, 5s.
+We exploit that and turn them into deltas relative to the previous cycles.
+Now we need deltas only for cycles where the sending rate changes, it's very cheap.
+This is what the contract actually stores: a mapping from cycle numbers to deltas.
```
NunoAlexandre, line 58:
```suggestion
We exploit that by storing them in the form of deltas relative to the previous cycles. 
That reduces storage costs since we only need to store the deltas for cycles where the sending rate changes.

This is, in fact, what the contract stores: a mapping from cycle numbers to deltas.
```

> a mapping from cycle numbers to deltas.

On this, isn't this map contextual? I.e, if we'd _only_ store that association, we couldn't figure the _from_ and _to_ of the stream. Could you expand on that or refresh my memory? :smile:

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+When the receiver wants to collect the funds, they can iterate over the finished cycles, reconstruct
+the added values from deltas and then calculate the collectable amount at this point in time.
+For any cycle before the first funding begins the added value is assumed to be 0.
+
+# The sender
+
+The sender has a balance and a set of parameters, which describe the sending process.
+
+The balance can be increased by topping up, which requires sending the assets to the contract.
+The currently unspent balance can be withdrawn and transferred out of the contract at any time.
```
NunoAlexandre, line 71:
```suggestion
The currently remaining balance can be withdrawn and transferred out of the contract at any time.
```

What is the difference between "withdrawn" and "transferred out"?

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Collecting
+
+When the receiver wants to collect the funds, they can iterate over the finished cycles, reconstruct
+the added values from deltas and then calculate the collectable amount at this point in time.
+For any cycle before the first funding begins the added value is assumed to be 0.
+
+# The sender
+
+The sender has a balance and a set of parameters, which describe the sending process.
```
NunoAlexandre, line 68:
```suggestion
The sender has an available balance, a per-block rate, and a set of receivers. From the available balance,
the per-block rate is disbursed across all the receivers on every block.
```

?

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+
+The balance can be increased by topping up, which requires sending the assets to the contract.
+The currently unspent balance can be withdrawn and transferred out of the contract at any time.
+
+The overall rate of funding of the sender is described with the amount per block.
+It limits the value, which can be taken from the balance on each block and sent to the receivers.
+This value stays constant over time unless explicitly updated.
```
NunoAlexandre, line 75:
```suggestion
The amount per block value defines the rate at which the funds are disbursed from the remaining balance 
to the receivers on each block. This value stays constant over time unless explicitly updated.
```

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+
+The sender maintains a list of receivers too, each of them with a weight.
+The weights regulate how the per-block amount is split between the receivers.
+For example, a receiver with weight 2 is going to get a share twice as big
+as a receiver with weight 1, but only half as big as another receiver with weight 4.
```
NunoAlexandre, line 80:
```suggestion
Finally, the sender maintains a list of receivers. 
Each receiver has an associated `weight`, which regulates how the amount per-block is 
split between the receivers.

For example, a receiver with weight 2 is going to get a share twice as big
as a receiver with weight 1, but only half as big as another receiver with weight 4.
```

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Starting sending
```
NunoAlexandre, line 82: I read [Starting Strength](https://startingstrength.com/) :laughing:

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+
+When the sender has a sufficient balance, a sending rate and a list of receivers, sending starts.
```
NunoAlexandre, line 84:
```suggestion
Given that a sender has sufficient balance, a sending rate per-block, and a list of receivers, sending starts automatically.
```

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+First, the funding period is calculated.
+Its start is the current block and its end is the block on which the balance will run out.
+Next, for each receiver the weighted share of the per-block rate is calculated.
+The receiver's deltas are updated to reflect that during the whole sending period on every block
+it's going to receive the calculated amount.
```
NunoAlexandre, line 89:
```suggestion
The receiver's deltas are updated to the calculated amount that will be disbursed every block throughout the whole sending period.
```

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+
+Let's take a look at an example of an application of a delta.
+The sender will be sending 1 per block or 5 per cycle.
+
+![](how_pool_works_4.png)
+
+The deltas are applied relative to the existing values.
+It doesn't matter if anybody else is funding the receiver, it won't affect this sender.
```
NunoAlexandre, line 97:
```suggestion
Note that this is a  timeline relative to a specific sender, therefore not considering possible other sends supporting this receiver.
```

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+
+Another important point is that the delta changes are usually split between two cycles.
+This reflects that the first cycle is only partially affected by the change in funding.
+Only the second one is fully affected and it must apply the rest of the delta.
```
NunoAlexandre, line 101: Very clear :ok_hand:

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+
+In this case, the total change of the per-cycle delta is +5 to start sending.
+The current cycle isn't fully affected though, only 2 out of 5 blocks are sending.
+It's effectively going to transfer only the amount of 2, which is reflected in the +2 delta change.
+On the other hand, the next cycle and the ones after it are going to transfer the full 5.
+This is expressed with the +3 delta change, which turns 2 per cycle into the full 5 per cycle.
```
NunoAlexandre, line 107:
```suggestion
It's effectively only going to transfer only the amount of 2 in this current cycle, which is reflected in the +2 delta change.
From the next block onwards, the amount of 5 will be disbursed. This is expressed with the +3 delta change, 
which turns the previous 2 per cycle into the full 5 per cycle.
```

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Stopping sending
```
NunoAlexandre, line 109: On this section, it would be more digestible to first present what we consider "stopping sending", and then explain the mechanisms through which we represent that. Maybe starting with what is now the last paragraph would already help a lot.

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+
+The old funding end deltas are reverted because they don't reflect the real funding end anymore.
+On the other hand, a new end is applied to the current block,
+just as if it was always supposed to be the end of the funding period.
+Now the receiver's future isn't affected by the sender anymore.
+The past stays untouched because the already sent funds are out of the sender's control.
+
+# The proxy
+
+The proxy is configured only with a list of weighted receivers.
```
NunoAlexandre, line 133:
```suggestion
The proxy is configured only with a list of receivers with an associated weight.
```

We have already mentioned above that a receiver has a weight. Do we need to repeat that here? Maybe yes, just wan to double-check the goal here.

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+It's not possible to send funds via a proxy which isn't configured.
```
NunoAlexandre, line 134: What is a non-configured proxy, or what would that look like?

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+The sum of the receivers' weights must always be a constant value,
```
NunoAlexandre, line 135: > weights must always be a constant value,

under which context? Not sure I understand? If I start with `[(a, 1), (b, 1)]` and then change it to `[(a, 1), (b, 2)]`, the sum of the weights is no longer the same as it was before. That's how I mean my question :)

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+which is defined in the contract and it's the same for all the proxies.
+
+Just like a receiver, the proxy has a mapping between the cycles and the deltas of received amounts.
```
NunoAlexandre, line 138:
```suggestion
Just like a receiver, the proxy has a mapping between the cycles and the deltas of the amounts to disburse.
```

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Sending via a proxy
+
+When a sender starts sending funds to a proxy, it does so in two steps.
+First it applies changes to the proxy's deltas similarly to how it would do with a receiver.
+Next, it iterates over all the proxy's receivers and applies changes to
+their deltas as if they were directly funded by the sender.
+The funding rate applied to the receivers is split according to their weights in the proxy.
+
+For the example sake let's assume that the delta's proxy weights sum must be equal 5.
```
NunoAlexandre, line 148:
```suggestion
For the example sake let's assume that the delta's proxy weights sum must be equal to 5.
```

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+The receivers get their shares, A with weight 3 gets 6 and B with weight 2 gets 4 per cycle.
+
+When a sender stops sending, the process is reversed like with regular receivers.
+All the deltas are once again applied, but with negative values.
+
+## Updating a proxy
+
+When the list of proxy receivers is updated, all funding must be moved to a new set of receivers.
+That's when the proxy's deltas come useful.
+For each cycle and each receiver, the proxy can tell, what total delta it has applied.
```
NunoAlexandre, line 165:
```suggestion
For each cycle and each receiver, the proxy can tell the total delta it has applied.
```

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+the receiver's weights and subtracted from the corresponding receiver's deltas.
+
+After removing its influence from one set of receivers the proxy must reapply itself on a new set.
+This is done in the same way as removal, but this time the deltas are added and not subtracted.
+
+### The current cycle problem
+
+Unlike the senders, the proxies store data with a per-cycle precision.
+When changing the set of receivers, a delta describing the current cycle may need to be applied.
+When it happens, it's unclear, what part of the per-cycle delta should be moved,
```
NunoAlexandre, line 181:
```suggestion
When it happens, it's unclear what part of the per-cycle delta should be moved,
```

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+because some funds were sent before the current block and some will be sent after it.
+
+![](how_pool_works_8.png)
+
+The solution is to ignore the problem and move the whole current cycle delta.
+Some funds already sent in the current cycle may disappear from one receiver and appear in another,
+but it's not dangerous, the receivers have no access to funds coming from an unfinished cycle.
```
NunoAlexandre, line 188:
```suggestion
Some funds already sent in the current cycle may disappear from one receiver and appear in another.
Such behaviour, however, is not of significant importance since the receivers have no access to funds 
coming from an unfinished cycle.
```

NunoAlexandre (commented):
Super cool to have this. Left some comments, mostly around language. I've also left some questions :)

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+The funding pool is a smart contract, which creates real-time streams of donations.
+One can start, alter or end the process of sending their funds at any time with immediate effect.
+The flow of funds is automatically maintained and steady over time.
+
+There are 3 roles present in the contract.
+Any Ethereum address can be used in any number of these roles at the same time.
+
+- The sender - has assets and chooses who do they want to send them to, how much and how fast
+- The receiver - receives funds from senders
+- The proxy - receives funds from senders, but immediately passes them to receivers of their choice
```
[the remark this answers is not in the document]
CodeSandwich: It means the speed of moving funds from the sender to the receiver

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+The whole blockchain history is divided into cycles of equal length.
+Every cycle contains the same, constant number of blocks.
+In the examples below, we assume that the cycle length is 5.
+
+![](how_pool_works_1.png)
+
+# The receiver
+
+The receiver requires no preparation, any address can receive donations at any time.
+The only function of this role is collection of funds sent by others.
```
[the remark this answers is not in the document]
CodeSandwich: How is this version better?

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+## The deltas
+
+On every block, a sender can send funds to a receiver.
+The receiver can collect funds sent on a given block only when the cycle containing it is over.
+
+![](how_pool_works_2.png)
+
+Here we have a history of a receiver, who has 2 senders.
+Each of them was sending different amounts over different periods of time.
+On the end of each cycle, the collectable amount was increased by the total sent amount.
```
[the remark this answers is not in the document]
CodeSandwich: How is this version better?

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+
+Each receiver needs to know, how much was sent to them.
+To accomplish that the history of received funds needs to be stored in some form.
+The receiver doesn't care, who exactly is sending on each block and in what configuration.
+The only thing that matters are the collectable per-cycle values below the timeline.
+This reduces the amount of stored data significantly, but this can be optimized further.
```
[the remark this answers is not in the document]
CodeSandwich: It's in the next paragraph

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+
+![](how_pool_works_3.png)
+
+In this example, we started with having the raw collectable value of 23 for every block until
+the end of the blockchain.
```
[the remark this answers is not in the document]
CodeSandwich: It's a description of data storage

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+We then reduced that to storing values added to the collectable amount on each cycle.
+Now we need to describe only cycles when receiving anything.
```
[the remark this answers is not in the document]
CodeSandwich: No, storing where changes take place is the next step. Have you seen the picture?

docs/how_pool_works.md
```diff
@@ -0,0 +1,189 @@
...
+The senders usually are sending constant per-cycle amounts over long periods of time, so
```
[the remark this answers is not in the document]
CodeSandwich: They can send over short periods too

docs/how_pool_works.md
CodeSandwich: What do you mean by contextual? The point is to NOT store the "from"

docs/how_pool_works.md
CodeSandwich: WIthdrawal is an action, the transfer is its underlying outcome, which results in moving funds out of the contract

docs/how_pool_works.md
CodeSandwich: This version is less precise

docs/how_pool_works.md
CodeSandwich: How is this better?

docs/how_pool_works.md
CodeSandwich: How is that version better?

docs/how_pool_works.md
CodeSandwich: That's not true!

docs/how_pool_works.md
CodeSandwich: It's a timeline of a receiver, not a sender, it considers all the senders

docs/how_pool_works.md
CodeSandwich: How is this version better?

docs/how_pool_works.md
CodeSandwich: Which paragraph do you mean?

docs/how_pool_works.md
CodeSandwich: We described above the sender's receivers, the proxy could have no weights, so it's important to mention

docs/how_pool_works.md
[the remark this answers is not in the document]
NunoAlexandre: I see. Isn't that describing things from a user point of view, which is sort of irrelevant here? I.e, we don't actually know for how long users will want to run their outgoing support, and as for the contracts, that information is not really important. What do you think?

docs/how_pool_works.md
NunoAlexandre: I mean that if there is no context and all we have is a cycle number and deltas, we'd have not enough data to actually perform a transfer.

docs/how_pool_works.md
NunoAlexandre: Don't quite get the point. It seems that this says it succinctly:

```suggestion
The currently remaining balance can be withdrawn from the contract at any time.
```

The definition of `withdraw` encapsulates it all it seems to me.

docs/how_pool_works.md
NunoAlexandre: sorry, my bad. updated.

docs/how_pool_works.md
NunoAlexandre: I mean the last paragraph of this section.

docs/how_pool_works.md
CodeSandwich: Fixed

docs/how_pool_works.md
CodeSandwich: You're right, if you try to make such change, you break the rule about the sum of weights, which is described here. There's no context, it's a global rule. What's your question?

docs/how_pool_works.md
CodeSandwich: The receiver doesn't disburse amounts, so that change would break this analogy

docs/how_pool_works.md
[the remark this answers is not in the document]
NunoAlexandre: I have, but can't understand what the goal fo this sentence is then.

docs/how_pool_works.md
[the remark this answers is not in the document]
NunoAlexandre: "The receiver requires no preparation,: is not clear. Preparation for what, by whom? What I suggest tries to make that clear. Moreover, it seems that the second bit of your first sentence and the second sentence merged and simplified communicate more concisely an clearly.

docs/how_pool_works.md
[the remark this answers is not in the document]
CodeSandwich: It's a rationalization of a storage design choice. This description is for the people reading the contract code, so I'd say, that it's important to have a proper mental model

docs/how_pool_works.md
CodeSandwich: That's the point, cycle numbers and deltas ARE enough to describe a transfer from the sender's perspective

docs/how_pool_works.md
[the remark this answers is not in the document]
CodeSandwich: Ok, but I'll leave the second sentence, it's important to mention the collection step, the funds don't appear in the wallet automatically. Is it ok?

docs/how_pool_works.md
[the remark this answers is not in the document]
CodeSandwich: We're going from collectable, through added to a delta shaving off data on each step. Skipping the middle step would be confusing

NunoAlexandre:
> How is this better?

I'm reviewing a piece of text, so if leave suggestions it's purely because I didn't find the
original text clear enough, with a fairly followable line of thought, or with a fair degree of readability.
I am unsure how to reply to this question in cases where the improvement appears obvious to me, such as this example:

``` diff
- On the other hand, the next cycle and the ones after it ...
+ From the next cycle onward, ...
```

docs/how_pool_works.md
CodeSandwich: But there's still only one timeline per receiver, the only relative thing are the changes, but that's already said in the previous sentence

docs/how_pool_works.md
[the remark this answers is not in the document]
NunoAlexandre: The original text as misplaces commas and the writing style sound is as if you are actively presenting this, while that is not how it's read. Pus, a receiver doesn't "have" senders, they _receive_ funds (from senders).

docs/how_pool_works.md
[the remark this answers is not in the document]
NunoAlexandre: exactly, I don't get what you mean by "speed"? As in, from the app perspective, I set how much I want to disburse "per month" across the receivers I have defined, not how fast the disbursement takes place. Could you elucidate me here? :)

docs/how_pool_works.md
CodeSandwich: Fixed

CodeSandwich:
Sorry about the vague questions, but I sometimes feel like the changes are proposed for the sake of introducing changes. Even worse they sometimes obscure the meaning by removing some details. It's a technical text, it must be first and foremost precise, the style is of secondary importance.

docs/how_pool_works.md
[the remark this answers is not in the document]
CodeSandwich: Do you have an idea of what could be a substitute? `and how fast` => `and with what rate`? But then again what's a rate? The concept of speed seems understandable and precise enough for the fourth paragraph of the whole text, the details are laid out later.

docs/how_pool_works.md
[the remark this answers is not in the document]
NunoAlexandre: Sounds good, Igor :)

RE the second sentence, there is a typo/missing word:

``` diff
The only function of this role is **the** collection of funds sent by others.
```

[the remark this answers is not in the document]
CodeSandwich: Fixed

docs/how_pool_works.md
NunoAlexandre: I don't think this sentence implies that the receiver is the one who disburses, but that the proxy has a mapping similar to the receivers.

NunoAlexandre:
> Sorry about the vague questions, but I sometimes feel like the changes are proposed for the sake of introducing changes. 

Absolutely my last intention. Ideally, I wouldn't find the need to suggest anything at all. I am sorry that you feel that way.

> Even worse they sometimes obscure the meaning by removing some details. It's a technical text, it must be first and foremost precise, the style is of secondary importance.

Observing details is definitely and obviously not a trade-off I am consciously making. It can be a result of not knowing as much of the subject in question as yourself and of the original text not elucidating me enough, which might have been the trigger for the suggestion in the first place.

docs/how_pool_works.md
CodeSandwich: After thorough reading you're right, it means amounts to disburse INTO a receiver/proxy. Is disburse a better word here than receive?

docs/how_pool_works.md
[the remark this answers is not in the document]
NunoAlexandre: Is the "speed" here referring to how fast the pool balance will be emptied out? I think so, just want to clarify.
If so:

> Do you have an idea of what could be a substitute? `and how fast` => `and with what rate`?

I'd say `and at what rate`. What do you think?

docs/how_pool_works.md
NunoAlexandre: Cool :)

> Is disburse a better word here than receive?

Since the flow here is proxy -> receivers I'd say "disburse", yeah.

CodeSandwich: I was viewing it as sener -> receiver, it's more consistent with the overall narration

docs/how_pool_works.md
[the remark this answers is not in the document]
CodeSandwich: Ok, fixed

docs/how_pool_works.md
NunoAlexandre: Yes you are right :) "receive" then :100:

docs/how_pool_works.md
```diff
@@ -0,0 +1,198 @@
...
+Every cycle contains the same, constant number of blocks.
+In the examples below, we assume that the cycle length is 5.
+
+![](how_pool_works_1.png)
+
+# The receiver
+
+There no setup steps for one to become a receiver.
+Any address can receive donations at any time, from any sender.
+The only function of this role is collection of funds sent by others.
```
NunoAlexandre, line 32:
```suggestion
The only function of this role is the collection of funds sent by others.
```

NunoAlexandre (commented):
Looking great, Igor :star: 

Could you pass the raw file against [grammarly.com](https://www.grammarly.com/) to fix all the remaining mistakes? It'd be faster that way and it gives you insights on why for every correctness alert and you can commit it all together.

docs/how_pool_works.md
CodeSandwich: I did pass it through Grammarly before submitting the PR and fixed most of the issues. I remember this exact issue and I remember explicitly NOT fixing it. Grammarly treated `collection` as a regular noun, but here it's a noun derived from a verb, which is too abstract to use `the` here.

NunoAlexandre: Not sure if it makes a difference but I use a paid version and there are currently 21 correctness alerts left, excluding this one.

CodeSandwich: I've used the free version, which mostly checks the grammar. The premium features are mostly about style, which may or may not be even relevant here. I got 69 issues in total, but with access to details about only 11.

docs/how_pool_works.md
```diff
@@ -0,0 +1,198 @@
...
+
+The whole blockchain history is divided into cycles of equal length.
+Every cycle contains the same, constant number of blocks.
+In the examples below, we assume that the cycle length is 5.
+
+![](how_pool_works_1.png)
+
+# The receiver
+
+There no setup steps for one to become a receiver.
```
cloudhead, line 30:
```suggestion
There are no setup steps for one to become a receiver.
```

docs/how_pool_works.md
cloudhead: "End of time" might be clearer because to me, the end of the chain is the head/tip :stuck_out_tongue:

CodeSandwich: Fixed

docs/how_pool_works.md
```diff
@@ -0,0 +1,198 @@
...
+From the available balance, the per-block rate is disbursed across all the receivers on every block.
+
+The balance can be increased by topping up, which requires sending the assets to the contract.
+The currently remaining balance can be withdrawn from the contract at any time.
+
+The overall rate of funding of the sender is described with the amount per block.
+It limits the value, which can be taken from the balance on each block and sent to the receivers.
+This value stays constant over time unless explicitly updated.
+
+The sender maintains a list of receivers too, each of them with a weight.
```
geigerzaehler, line 81: Why “too”? Does this reference another thing that maintains a list of receivers? We’re already mentioning receivers at the beginning of the section so I don’t think we need the “too”.

docs/how_pool_works.md
```diff
@@ -0,0 +1,198 @@
...
+When the receiver wants to collect the funds, they can iterate over the finished cycles, reconstruct
+the added values from deltas and then calculate the collectable amount at this point in time.
+For any cycle before the first funding begins the added value is assumed to be 0.
+
+# The sender
+
+The sender has an available balance, a per-block rate, and a set of receivers.
+From the available balance, the per-block rate is disbursed across all the receivers on every block.
+
+The balance can be increased by topping up, which requires sending the assets to the contract.
```
geigerzaehler, line 74: We should clearly point out the actions that a sender can take and mention that explicitly instead of using passive voice.

docs/how_pool_works.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Collecting
+
+When the receiver wants to collect the funds, they can iterate over the finished cycles, reconstruct
+the added values from deltas and then calculate the collectable amount at this point in time.
+For any cycle before the first funding begins the added value is assumed to be 0.
+
+# The sender
+
+The sender has an available balance, a per-block rate, and a set of receivers.
```
geigerzaehler, line 71: * We need to find the right terminology for the balance. Is “available” needed? What is the “unavailable” part of the balance. How does it relate to the “currently remaining balance” mentioned below?
* Instead of “per-block rate” I’d suggest the term “funding rate” and then explain that it is expressed in tokens (or funds or something else) per block.

docs/how_pool_works.md
```diff
@@ -0,0 +1,198 @@
...
+From the available balance, the per-block rate is disbursed across all the receivers on every block.
```
geigerzaehler, line 72: We should make the relation ship between the balance an  rate clearer: The balance decreases every block by the funding rate. The funds are credited to the receivers. I’d also avoid the term “disbursed ... on every block” because it sounds like this is an explicit action and one might have the impression that this changes something in the chain state while it is only a bookkeeping thing.

docs/how_pool_works.md
```diff
@@ -0,0 +1,198 @@
...
+# The receiver
+
+There are no setup steps for one to become a receiver.
+Any address can receive donations at any time, from any sender.
+The only function of this role is collection of funds sent by others.
+
+## The deltas
+
+On every block, a sender can send funds to a receiver.
+The receiver can collect funds sent on a given block only when the cycle containing it is over.
```
geigerzaehler, line 37: We should elaborate on this
* The “collect” transaction triggered by a receiver moves all collectable funds to the recipients account
* We need to explain how the collectable funds are calculated
* We need to explain that collecting funds reduces the collectable funds.

docs/how_pool_works.md
geigerzaehler: Without “the” the sentence sound off to me.

geigerzaehler (commented):
Did an initial pass. The pictures are a great help.

docs/how_pool_works.md
CodeSandwich: Fixed

docs/how_pool_works.md
```diff
@@ -0,0 +1,198 @@
...
+
+- **The sender**: has assets and chooses who do they want to send them to,
+how much, and at what rate
+- **The receiver:** receives funds from senders
+- **The proxy**: receives funds from senders,
+but immediately passes them to receivers of their choice
+
+## The cycles
+
+The whole blockchain history is divided into cycles of equal length.
```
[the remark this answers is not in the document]
CodeSandwich: Fixed

docs/how_pool_works.md
```diff
@@ -0,0 +1,198 @@
...
+Every cycle contains the same, constant number of blocks.
```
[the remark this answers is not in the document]
CodeSandwich: Fixed

docs/how_pool_works.md
CodeSandwich: Fixed

docs/how_pool_works.md
CodeSandwich: Fixed

docs/how_pool_works.md
CodeSandwich: Both fixed

docs/how_pool_works.md
CodeSandwich: Fixed

docs/how_pool_works.md
```diff
@@ -0,0 +1,198 @@
...
+Now we need deltas only for cycles where the sending rate changes, it's very cheap.
+This is what the contract actually stores: a mapping from cycle numbers to deltas.
+
+## Collecting
+
+When the receiver wants to collect the funds, they can iterate over the finished cycles, reconstruct
+the added values from deltas and then calculate the collectable amount at this point in time.
+For any cycle before the first funding begins the added value is assumed to be 0.
+
+# The sender
```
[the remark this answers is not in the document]
CodeSandwich: Fixed

geigerzaehler (approved):
This is a great start already :sparkles:! But I feel like this still needs some iterations until people that are not familiar with the concepts already can understand this. Instead of trying to achieve this in this PR I would propose we do this in future iterations. Feel free to address the issues you deem important for this to be merged, @CodeSandwich.

docs/how_pool_works.md
```diff
@@ -0,0 +1,225 @@
...
+The senders usually are sending constant per-cycle amounts over long periods of time, so
+the added values tend to create long series of constant numbers, in this case, 5s.
+We exploit that and finally turn them into **deltas** relative to the previous cycles.
+Now we need to store data only for cycles where the sending rate changes, it's very cheap.
+This is what the contract actually stores: a mapping from cycle numbers to deltas.
+
+## Starting sending
+
+When the sender has a sufficient balance, a sending rate and a list of receivers,
+the flow of assets automatically starts.
```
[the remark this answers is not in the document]
CodeSandwich: Fixed

docs/how_pool_works.md
```diff
@@ -0,0 +1,225 @@
...
+Now we need to describe only cycles when receiving anything.
+The senders usually are sending constant per-cycle amounts over long periods of time, so
+the added values tend to create long series of constant numbers, in this case, 5s.
+We exploit that and finally turn them into **deltas** relative to the previous cycles.
+Now we need to store data only for cycles where the sending rate changes, it's very cheap.
+This is what the contract actually stores: a mapping from cycle numbers to deltas.
+
+## Starting sending
+
+When the sender has a sufficient balance, a sending rate and a list of receivers,
```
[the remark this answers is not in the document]
CodeSandwich: All fixed

docs/how_pool_works.md
```diff
@@ -0,0 +1,225 @@
...
+These amounts are then added up and the result is the total collected amount.
+Finally, the next uncollected cycle number and the last cycle amount are updated.
+
+![](how_pool_works_6.png)
+
+In this example funds received from 4 cycles are being collected.
+The yellow fields are the stored state before the collection, green after it.
+The blue field is the collected value, which is going to be transferred to the sender's wallet.
+
+# The proxy
```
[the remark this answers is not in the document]
CodeSandwich: Added

docs/how_pool_works.md
```diff
@@ -0,0 +1,225 @@
...
+which results in sending the assets from the contract back to the user wallet.
+
+This funding rate stays constant over time unless explicitly updated.
+
+The sender maintains a list of receivers, each of them with a weight.
+The weights regulate how the funded amount is split between the receivers.
+For example, a receiver with weight 2 is going to get a share twice as big
+as a receiver with weight 1, but only half as big as another receiver with weight 4.
+
+## The deltas
```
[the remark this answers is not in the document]
CodeSandwich: Agreed, it's cluttered and murky. Applied your proposition with some changes to fit the surrounding text.

docs/how_pool_works.md
```diff
@@ -0,0 +1,225 @@
...
+
+## Collecting
+
+The receiver can at any time collect the funds sent to it.
+The contract calculates the total amount and then transfers it out to the receiver's wallet.
+The collected amount is always everything available at a given moment, there's no way to limit it.
+
+As shown in the previous sections, the collectable amount is described with deltas, one per cycle.
+The receiver stores the number of the first cycle, for which the funds haven't been collected yet.
+This prevents collecting the same funds multiple times, every cycle can be ever processed only once.
```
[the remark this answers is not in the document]
CodeSandwich: Good point, fixed

docs/how_pool_works.md
```diff
@@ -0,0 +1,225 @@
...
+
+When a sender stops sending, the process is reversed like with regular receivers.
+All the deltas are once again applied, but with negative values.
+
+## Updating a proxy
+
+When the list of proxy receivers is updated, all funding must be moved to a new set of receivers.
+That's when the proxy's deltas come useful.
+For each cycle and each receiver, the proxy can tell the total delta it has applied.
+It can then use this information to erase its influence from its receivers.
```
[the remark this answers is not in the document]
CodeSandwich: Fair, fixed

docs/how_pool_works.md
```diff
@@ -0,0 +1,225 @@
...
+In the examples below, we assume that the cycle length is 5.
+
+![](how_pool_works_1.png)
+
+# The sender
+
+The sender has a balance, a funding rate, and a set of receivers.
+
+The balance is automatically reduced by the funding rate on every block
+and the same amount is credited to the sender's receivers.
```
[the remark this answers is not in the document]
CodeSandwich: Fixed

docs/how_pool_works.md
```diff
@@ -0,0 +1,225 @@
...
+This funding rate stays constant over time unless explicitly updated.
+
+The sender maintains a list of receivers, each of them with a weight.
+The weights regulate how the funded amount is split between the receivers.
+For example, a receiver with weight 2 is going to get a share twice as big
+as a receiver with weight 1, but only half as big as another receiver with weight 4.
+
+## The deltas
+
+On every block, a sender sends funds to a receiver.
```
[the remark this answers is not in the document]
CodeSandwich: Applied

README.md
```diff
@@ -2,6 +2,9 @@
 
 See [`DEVELOPING.md`](./DEVELOPING.md) for the developer manual.
 
+See [`how_pool_works.md`](./docs/how_pool_works.md) for the introduction to
```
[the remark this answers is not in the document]
CodeSandwich: Fixed

NunoAlexandre (approved):

PR #61, opened 2020-11-25
Unify test tools for ETH and ERC-20 pools

CodeSandwich:
A clean up and dedupllication of the test utils

### DIFF
--- a/test/pool.test.ts
+++ b/test/pool.test.ts
@@ -7,7 +7,7 @@ import { Erc20 } from "../contract-bindings/ethers/Erc20";
 import { Erc20Pool } from "../contract-bindings/ethers/Erc20Pool";
 import { EthPool } from "../contract-bindings/ethers/EthPool";
 import { ethers } from "hardhat";
-import { Signer, BigNumber } from "ethers";
+import { Signer, BigNumber, ContractTransaction } from "ethers";
 import { expect } from "chai";
 import {
   randomAddress,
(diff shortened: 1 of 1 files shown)


test/pool.test.ts
```diff
@@ -28,6 +28,20 @@ async function mineBlocksUntilCycleEnd(): Promise<void> {
...
 type AnyPool = EthPool | Erc20Pool;
 
+type ReceiverWeights = [PoolUser, number][];
+
+function receiverWeightsForContract(
+  weights: ReceiverWeights
+): {
+  receiver: string;
+  weight: number;
+}[] {
```
geigerzaehler, line 38: Just a stylistic suggestion.
```suggestion
): Array<{
  receiver: string;
  weight: number;
}> {
```

For me, this is easier to parse since the important information is at the start of the type and not as easy to miss as `[]` at the end. I usually choose this syntax whenever it is more complicated than `SomeType[]`.

test/pool.test.ts
```diff
@@ -72,6 +86,51 @@ abstract class PoolUser {
...
+  async collect(expectedAmount: number): Promise<void> {
+    await this.expectCollectableOnNextBlock(expectedAmount);
+    const balanceBefore = await this.getBalance();
+    await submit(this.pool.collect({ gasPrice: 0 }), "collect");
+    const balanceAfter = await this.getBalance();
+    const collected = balanceAfter.sub(balanceBefore).toNumber();
+    expect(collected).to.equal(
+      expectedAmount,
+      "The collected amount is different from the expected amount"
+    );
```
geigerzaehler, line 102: Making assertions on a balance change seems to be a common pattern. How about this.
```suggestion
    await this.expectBalanceChange(
      () => {
        await submit(this.pool.collect({ gasPrice: 0 }), "collect");
      },
      expectedAmount,
      "The collected amount is different from the expected amount"
    );
```

geigerzaehler (approved):

test/pool.test.ts
CodeSandwich: Great catch, applied

test/pool.test.ts
CodeSandwich: I agree, fixed

PR #69, opened 2020-12-10
Batching

### LINKED ISSUE #50: Add batch sender update
The sender has to make at least 3 steps to start donating: set a per month rate, top up the account and configure a list of receivers, which itself can take up to tens of transactions. This is a slow and expensive process.

We need a batch update transaction, which can take all these parameters and set them all at once.

CodeSandwich:
Fixes #50 .

The proposed API may seem weird with 2 separate parameters for top up and withdrawal. This is necessary for the symmetry between the ETH and ERC-20 APIs. While the latter could get away with having a single parameter with a sign, the former forces top up to be a separate value passed as a transaction value.

The inner structure of the pool contract is a little too layered due to the `suspendPayments` modifier.

I think that there's no need to have two APIs doing the same things for the same cost. I think that we should drop the old API for altering separate sending parameters. This will significantly simplify code.

### DIFF
--- a/contracts/Pool.sol
+++ b/contracts/Pool.sol
@@ -192,13 +192,13 @@ abstract contract Pool {
         }
         receiver.lastFundsPerCycle = lastFundsPerCycle;
         receiver.nextCollectedCycle = collectedCycle;
-        if (collected > 0) transferToSender(uint128(collected));
+        transferToSender(uint128(collected));
     }
 
     /// @notice Must be called when funds have been transferred into the pool contract
     /// in order to top up the message sender
     /// @param amount The topped up amount
-    function onTopUp(uint128 amount) internal suspendPayments {
+    function topUpSuspending(uint128 amount) internal suspendPayments {
         senders[msg.sender].startBalance += amount;
     }
 
@@ -435,7 +435,7 @@ abstract contract Pool {
     }
 
     /// @notice Called when funds need to be transferred out of the pool to the message sender
-    /// @param amount The transferred amount, never zero
+    /// @param amount The transferred amount
     function transferToSender(uint128 amount) internal virtual;
 
     /// @notice Stops payments of `msg.sender` for the duration of the modified function.
@@ -684,11 +684,11 @@ contract EthPool is Pool {
 
     /// @notice Tops up the sender balance of a sender of the message with the amount in the message
     function topUp() public payable {
-        if (msg.value > 0) onTopUp(uint128(msg.value));
+        if (msg.value > 0) topUpSuspending(uint128(msg.value));
     }
 
     function transferToSender(uint128 amount) internal override {
-        msg.sender.transfer(amount);
+        if (amount != 0) msg.sender.transfer(amount);
     }
 }
 
@@ -713,11 +713,15 @@ contract Erc20Pool is Pool {
     /// @param amount The amount to top up with
     function topUp(uint128 amount) public {
         if (amount == 0) return;
-        erc20.transferFrom(msg.sender, address(this), amount);
-        onTopUp(amount);
+        transferToContract(amount);
+        topUpSuspending(amount);
+    }
+
+    function transferToContract(uint128 amount) internal {
+        if (amount != 0) erc20.transferFrom(msg.sender, address(this), amount);
     }
 
     function transferToSender(uint128 amount) internal override {
-        erc20.transfer(msg.sender, amount);
+        if (amount != 0) erc20.transfer(msg.sender, amount);
     }
 }
--- a/contracts/Pool.sol
+++ b/contracts/Pool.sol
@@ -76,7 +76,7 @@ abstract contract Pool {
     /// Whenever a proxy is added to a sender, this number is added to its `weightCount`.
     /// Limits costs of changes in sender's or proxy's configuration.
     uint32 public constant PROXY_WEIGHTS_COUNT_MAX = 10;
-    /// @notice The amount passed to `withdraw` to withdraw all the funds
+    /// @notice The amount passed as the withdraw amount to withdraw all the funds
     uint128 public constant WITHDRAW_ALL = type(uint128).max;
     /// @notice The amount passed as the amount per block to keep the parameter unchanged
     uint128 public constant AMOUNT_PER_BLOCK_UNCHANGED = type(uint128).max;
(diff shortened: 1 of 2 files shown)


contracts/Pool.sol
```diff
@@ -272,6 +342,22 @@ abstract contract Pool {
...
+    /// Setting a non-zero weight for a new receiver or
+    /// a new proxy adds it to the list of the sender's receivers.
+    /// Setting zero as the weight for a receiver or a proxy
+    /// removes it from the list of the sender's receivers.
+    /// @param updatedReceivers The list of the updated receivers and their new weights
+    /// @param updatedProxies The list of the updated proxies and their new weights
+    function setReceiversInternal(
+        ReceiverWeight[] calldata updatedReceivers,
+        ReceiverWeight[] calldata updatedProxies
+    ) public {
```
geigerzaehler, line 360: Should this really be public?

contracts/Pool.sol
```diff
@@ -76,8 +76,10 @@ abstract contract Pool {
     /// Whenever a proxy is added to a sender, this number is added to its `weightCount`.
     /// Limits costs of changes in sender's or proxy's configuration.
     uint32 public constant PROXY_WEIGHTS_COUNT_MAX = 10;
-    /// @notice The amount passed to `withdraw` to withdraw all the funds
+    /// @notice The amount passed as the withdrawn amount to withdraw all the funds
```
geigerzaehler, line 79: Removing the “n” at the end aligns it with the parameter name (`withdrawAmt`) and makes it clearer that this controls the amount to be withdrawn and not the amount that has actually be withdrawn, that is the return value `withdrawn`.

```suggestion
    /// @notice The amount passed as the withdraw amount to withdraw all the funds
```

contracts/Pool.sol
```diff
@@ -190,14 +192,58 @@ abstract contract Pool {
...
+        setAmountPerBlockInternal(amountPerBlock);
+        setReceiversInternal(updatedReceivers, updatedProxies);
     }
 
     /// @notice Must be called when funds have been transferred into the pool contract
     /// in order to top up the message sender
     /// @param amount The topped up amount
-    function onTopUp(uint128 amount) internal suspendPayments {
-        senders[msg.sender].startBalance += amount;
+    function topUpSuspending(uint128 amount) internal suspendPayments {
```
geigerzaehler, line 238: Could you update the documentation for this function?

contracts/Pool.sol
```diff
@@ -711,11 +839,60 @@ contract Erc20Pool is Pool {
...
+    /// a new proxy adds it to the list of the sender's receivers.
+    /// Setting zero as the weight for a receiver or a proxy
+    /// removes it from the list of the sender's receivers.
+    /// @param topUpAmt The topped up amount
+    /// @param withdraw The amount to be withdrawn, must not be higher than available funds.
+    /// Can be `WITHDRAW_ALL` to withdraw everything.
+    /// @param amountPerBlock The target amount to be sent on every block.
+    /// Can be `AMOUNT_PER_BLOCK_UNCHANGED` to keep the amount unchanged.
+    /// @param updatedReceivers The list of the updated receivers and their new weights
+    /// @param updatedProxies The list of the updated proxies and their new weights
```
geigerzaehler, line 871: I’m totally fine referencing the documentation of `updateSenderInternal` here. This avoids duplicating the documentation and it getting out of sync.

test/pool.test.ts
```diff
@@ -1119,6 +1219,77 @@ describe("EthPool", function () {
...
+    await proxy.setProxyWeights([[receiver2, proxyWeightBase]]);
+    await sender.updateSender(
+      0,
+      proxyWeightBase * 6,
+      proxyWeightBase * 3,
+      [[receiver1, proxyWeightBase]],
+      [[proxy, proxyWeightBase * 2]]
+    );
+    await mineBlocks(1);
+    await mineBlocksUntilCycleEnd();
```
geigerzaehler, line 1235: Why do we have two mine statements? Can we add a comment for this or somehow abstract it away?

test/pool.test.ts
```diff
@@ -1136,4 +1307,75 @@ describe("Erc20Pool", function () {
     await mineBlocksUntilCycleEnd();
     await receiver.collect(10);
   });
+
+  it("Allows full sender update with top up", async function () {
```
geigerzaehler, line 1311: One thing we might want to consider for the future is to abstract the test cases over the concrete pool implementation. Meaning that we would write the test cases once for an abstract pool and run them against both the ERC20 and the Eth pool.

geigerzaehler (commented):
:money_with_wings:

test/pool.test.ts
CodeSandwich: Good catch, removed. It makes no sense since `mineBlocksUntilCycleEnd` always mines at least 1 block.

contracts/Pool.sol
CodeSandwich: I'm not sure if it's worth deduplication. There are some issues with it:

- `updateSenderInternal` is `internal`, will its doc strings be easily accessible to the consumers?
- `updateSenderInternal` can't say that it sends any funds, it returns the amount which should be sent
- Different `updateSender`s handle incoming funds slightly differently which leads to different lists of parameters (message amount) and different prerequisites (allowance)

I'm afraid that we'll turn 3 self-containing texts into a web of references, amendments and conditions.

geigerzaehler: I don’t meant that we *just* reference `updateSenderInternal`. We should definitely include the differences. But for the parameter documentation for example and the rest that is shared we can just refer to `updateSenderInternal`.

> `updateSenderInternal` is `internal`, will its doc strings be easily accessible to the consumers?

At the moment we don’t produce any output and the documentation is only consumed by reading the comments.

This is just a suggestion though. If you prefer, we can keep it as it is.

test/pool.test.ts
CodeSandwich: Brilliant idea worth exploring!

contracts/Pool.sol
CodeSandwich: What changes would you expect to see? The function didn't change its behavior or API except the name.

contracts/Pool.sol
CodeSandwich: Fixed

contracts/Pool.sol
CodeSandwich: Great catch! Absolutely not, it's a critical bug :scream: I've reviewed all the public functions and found `setAmountPerBlockInternal` with the same issue.

contracts/Pool.sol
CodeSandwich: I'd prefer leaving it as it is. I to don't like the redundancy, but forcing users to mental stitching of the function docs is IMO even worse.

contracts/Pool.sol
geigerzaehler: I just find the “must be called” portion confusing. Maybe just plainly describing what it does:

> Adds the given amount to the senders pool balance and recalculates the funding period.

CodeSandwich: Fixed

NunoAlexandre (approved):
Looks good to me and it all seems to be working nicely in the app :smiley: 

Nothing to add on top of Thomas' comments.

geigerzaehler (approved):

Issue #71, opened 2020-12-10
Create governance system


cloudhead:
To be able to update our contracts, we need a governance system that uses the radicle token to vote. We're going to be forking the compound governance modules for this, and the COMP token.

* [x] Create governor contract (#72)
* [x] Create timelock contract (#72)
* [x] Update token contract
* [x] Create proxy contract
* [ ] ~~Set correct governance parameters for Radicle~~

PR #72, opened 2020-12-11
Import and adapt Compound Governor

cloudhead:
This is the first of several changes to address #71. Here, we just import the compound governor and token, and adapt them so that they compile.

I've conserved the original contract, to keep track of the changes I've made (c57577d).

You can find the originals here: https://github.com/compound-finance/compound-protocol/tree/master/contracts/Governance


contracts/Governance/SafeMath.sol
```diff
@@ -0,0 +1,185 @@
+pragma solidity ^0.7.5;
```
CodeSandwich, line 1: Don't we already have `contracts/libraries/SafeMath.sol`? :thinking:

cloudhead: Hah, good catch.

contracts/Governance/Token.sol
```diff
@@ -0,0 +1,391 @@
...
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+pragma solidity ^0.7.5;
+pragma experimental ABIEncoderV2;
+
+contract Token {
```
geigerzaehler, line 30: Could we prefix this with `Governance`? Otherwise it shows up as generic `Token` everywhere else.

cloudhead: I will rename to `RadicleToken`. Then, I will remove `Rad.sol` and point everything to `RadicleToken.sol`.

CodeSandwich (approved):
LGTM!

Issue #67, opened 2020-12-03
Contract updates


CodeSandwich:
We need a way to update the contracts. This consists of 2 problems:
- Swapping out the contract itself
- Preserving the storage

cloudhead:
Closing, as we have the proxy contract now.

PR #73, opened 2020-12-17
Governance proxy contracts

cloudhead:
Adds a proxy contract that is basically Compound's "Unitroller" contract.

We can use this as a template to make some of our contracts upgradable.

Related to #67


contracts/Proxy.sol
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    /**
+     * @notice Accepts new implementation. `msg.sender` must be pendingImplementation
+     * @dev Admin function for new implementation to accept its role as implementation
+     * @return uint 0=success, otherwise a failure (see ErrorReporter.sol for details)
+     */
+    function _acceptImplementation() public returns (uint256) {
+        // Check caller is pendingImplementation and pendingImplementation ≠ address(0)
+        if (msg.sender != pendingImplementation || pendingImplementation == address(0)) {
```
CodeSandwich, line 72: Can msg.sender be ever 0?

CodeSandwich (approved):
Very clever, entertaining to review :D

contracts/Proxy.sol
```diff
@@ -0,0 +1,187 @@
...
+            return
+                fail(Error.UNAUTHORIZED, FailureInfo.ACCEPT_PENDING_IMPLEMENTATION_ADDRESS_CHECK);
+        }
+
+        // Save current values for inclusion in log
+        address oldImplementation = implementation;
+        address oldPendingImplementation = pendingImplementation;
+
+        implementation = pendingImplementation;
```
CodeSandwich, line 81: Shouldn't we call `pendingImplementation.upgrade(self)`? Otherwise it doesn't have a right to veto the upgrade and a chance to perform any migrations.

contracts/Proxy.sol
```diff
@@ -0,0 +1,187 @@
...
+     */
+    event NewAdmin(address oldAdmin, address newAdmin);
+
+    constructor() {
+        // Set admin to caller
+        admin = msg.sender;
+    }
+
+    /*** Admin Functions ***/
+    function _setPendingImplementation(address newPendingImplementation) public returns (uint256) {
```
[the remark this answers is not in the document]
cloudhead: I think this allows users to handle the error if they want to, vs. the tx simply being reverted, which isn't a bad feature. I'm not sure why exactly they (Compound) went with this design, but I didn't want to change it too much either.

contracts/Proxy.sol
cloudhead: It can't, no. But this doesn't check that, it checks if any of these states is true and exits:

* the method was called from the wrong address
* the pending implementation hasn't been set yet

contracts/Proxy.sol
cloudhead: The upgrade process is started from this contract (`_setPendingImplementation`), but completed from the contract being upgraded. So `_acceptImplementation` is called by the target contract, after it has prepared for an upgrade. See here: https://github.com/radicle-dev/radicle-contracts/pull/73/files#diff-822ef81b3cb234d98d0343ec37a053b9c6a62a021b027e87e2f4e8cb646180d6R10

contracts/ErrorReporter.sol
```diff
@@ -0,0 +1,66 @@
...
+        ACCEPT_PENDING_IMPLEMENTATION_ADDRESS_CHECK,
+        SET_PENDING_ADMIN_OWNER_CHECK,
+        SET_PENDING_IMPLEMENTATION_OWNER_CHECK
+    }
+
+    /**
+     * @dev `error` corresponds to enum Error; `info` corresponds to enum FailureInfo, and `detail` is an arbitrary
+     * contract-specific code that enables us to report opaque error codes from upgradeable contracts.
+     **/
+    event Failure(uint256 error, uint256 info, uint256 detail);
```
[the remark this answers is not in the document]
cloudhead: Could it be a remnant from solc 0.5.*?

PR #75, opened 2020-12-17
Cheaperators

CodeSandwich:
Cuts 20-33 % of gas cost of iterating over mappings. The larger the mapping the bigger the saving.

Unfortunately it makes the iteration API even more complex. I've tried wrapping the swarm of parameters in a structure, but that forced it to be stored in memory leading to a considerable cost increase.

### DIFF
--- a/contracts/libraries/ProxyDeltas.sol
+++ b/contracts/libraries/ProxyDeltas.sol
@@ -15,68 +15,63 @@ struct ProxyDeltas {
 /// The list works optimally if after applying a series of changes it's iterated over.
 /// The list uses 2 words of storage per stored cycle.
 library ProxyDeltasImpl {
+    using ProxyDeltasImpl for ProxyDeltas;
+
     struct ProxyDeltaStored {
         uint64 next;
         int128 thisCycleDelta;
+        bool isAttached;
+        // Unused. Hints the compiler that it has full control over the content
+        // of the whole storage slot and allows it to optimize more aggressively.
+        uint56 slotFiller1;
         // --- SLOT BOUNDARY
         int128 nextCycleDelta;
+        // Unused. Hints the compiler that it has full control over the content
+        // of the whole storage slot and allows it to optimize more aggressively.
+        uint128 slotFiller2;
     }
 
     uint64 internal constant CYCLE_ROOT = 0;
-    uint64 internal constant CYCLE_END = type(uint64).max;
 
     /// @notice Return the next non-zero, non-obsolete delta and its cycle.
     /// The order is undefined, it may or may not be chronological.
     /// Prunes all the fully zeroed or obsolete items found between the current and the next cycle.
     /// Iterating over the whole list prunes all the zeroed and obsolete items.
-    /// @param current The previously returned cycle or CYCLE_ROOT to start iterating
+    /// @param prevCycle The previously returned `cycle` or CYCLE_ROOT to start iterating
+    /// @param prevCycleHint The previously returned `cycleHint` or CYCLE_ROOT to start iterating
     /// @param finishedCycle The last finished cycle.
     /// Entries describing cycles before `finishedCycle` are considered obsolete.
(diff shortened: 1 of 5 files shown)


contracts/libraries/ProxyDeltas.sol
```diff
@@ -15,68 +15,58 @@ struct ProxyDeltas {
...
     /// The order is undefined, it may or may not be chronological.
     /// Prunes all the fully zeroed or obsolete items found between the current and the next cycle.
     /// Iterating over the whole list prunes all the zeroed and obsolete items.
-    /// @param current The previously returned cycle or CYCLE_ROOT to start iterating
+    /// @param oldCurrent The previously returned cycle or CYCLE_ROOT to start iterating
+    /// @param oldNext The previously returned `next` or CYCLE_ROOT to start iterating
     /// @param finishedCycle The last finished cycle.
     /// Entries describing cycles before `finishedCycle` are considered obsolete.
-    /// @return next The next iterated cycle or CYCLE_ROOT if the end of the list was reached.
+    /// @return current The next iterated cycle or CYCLE_ROOT if the end of the list was reached.
```
geigerzaehler, line 39: When I read through this and the code the generic terminology (old, current, next) made it harder to follow the logic. Especially here where it says “current: the nexter iterated cycle”.

To improve this I’d suggest the following signature
~~~solidity
function nextDeltaPruning(
    ProxyDeltas storage self,
    uint64 prevCycle,
    uint64 cycleHint,
    uint64 finishedCycle
)
    internal
    returns (
        uint64 cycle,
        uint64 cycleHint,
        int128 thisCycleDelta,
        int128 nextCycleDelta
    )
~~~

contracts/libraries/ProxyDeltas.sol
```diff
@@ -15,68 +15,58 @@ struct ProxyDeltas {
...
 library ProxyDeltasImpl {
+    using ProxyDeltasImpl for ProxyDeltas;
+
     struct ProxyDeltaStored {
         uint64 next;
         int128 thisCycleDelta;
+        uint64 isAttached;
         // --- SLOT BOUNDARY
         int128 nextCycleDelta;
+        uint128 slotFiller;
```
geigerzaehler, line 26: Can you explain why this can’t just be omitted?

contracts/libraries/ProxyDeltas.sol
```diff
@@ -89,16 +79,20 @@ library ProxyDeltasImpl {
         int128 thisCycleDeltaAdded,
         int128 nextCycleDeltaAdded
     ) internal {
-        require(cycle != CYCLE_ROOT && cycle != CYCLE_END, "Invalid cycle number");
-        // Item not attached to the list
-        if (self.data[cycle].next == CYCLE_ROOT) {
```
geigerzaehler, line 94: Could you elaborate why using `isAttached` is cheaper? Seems like we need to store more now, no?

CodeSandwich: EVM reads and writes the whole slots anyway so slipping in another variable is practically free. We're greatly simplifying logic, which is good not only for humans, but it also lets the compiler shave off a few hundred gas.

contracts/libraries/ProxyDeltas.sol
CodeSandwich: Added a comment. The unused parts of the slot are assumed to be out of control and worth preserving. When a non-full slot is deleted, Solidity will read it, zero out the fields it manages and write back whatever is left. AFAIK it's supposed to let you do some magic with the empty parts from Yul and manage them yourself. Personally I'm not convinced that it's a useful design decision, for 99 % of contracts it's a nasty trap, but that's the way it is.

contracts/libraries/ProxyDeltas.sol
CodeSandwich: Great idea with the `Hint` convention, applied

geigerzaehler (approved):
Nice

PR #80, opened 2021-01-04
Update VRad contract to point to gov. token

cloudhead:
Just updated VRad for the new token contract.


CodeSandwich (approved):
At this point do we still need `Rad.sol`?

cloudhead:
> At this point do we still need `Rad.sol`?

We don't, we can slowly remove from the tests and stuff if we want, or maybe rename it `DummyToken`?

PR #82, opened 2021-01-06
Move @uniswap and @ensdomains to "dependencies"

NunoAlexandre:
These are not available in the exported build that clients use, requiring
"all" clients to add these exact dependencies themselves in order to run
the 'deploy.ts' script. That should happen here instead to that it can
actually offer the functionality it proposes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,9 +4,11 @@
   "license": "GPL-3.0-only",
   "main": "build/src/index.js",
   "types": "build/src/index.d.ts",
-  "dependencies": {},
-  "devDependencies": {
+  "dependencies": {
     "@ensdomains/ens": "^0.4.5",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -47,4 +48,4 @@
     "build/**",
     "contract-bindings/**"
   ]
-}
+}
```
geigerzaehler, line 51: Could you keep the newline to avoid unnecessary diff noise?

geigerzaehler (approved):

package.json
NunoAlexandre, line 51:
```suggestion
}

```

NunoAlexandre: That was unintentional, sorry. Will check my editor :eyes:

Fixed btw

PR #84, opened 2021-01-06
New vesting contract

cloudhead:
Replaces the old vesting contract with a new one, which is deployed once per beneficiary. It's less likely to be buggy as it's in use, and the logic is simpler. It was developed by Melonport.

Part of #88


contracts/Governance/VestingToken.sol
```diff
@@ -0,0 +1,119 @@
...
+    /// @notice Grant vesting tokens to the beneficiary.
+    ///
+    /// @param ofBeneficiary Address of beneficiary
+    /// @param ofTokenAmount Amount of tokens
+    /// @param ofVestingPeriod Vesting period in seconds from vestingStartTime
+    function grantTokens(
+        address ofBeneficiary,
+        uint256 ofTokenAmount,
+        uint256 ofVestingPeriod
+    ) external onlyOwner notInterrupted {
```
CodeSandwich, line 71: I think that moving these parameters to the constructor would simplify logic and reduce storage access. The vested contract has funds granted only once in its lifetime anyway and it doesn't make sense to have it constructed and not granted.

contracts/Governance/VestingToken.sol
```diff
@@ -0,0 +1,119 @@
...
+/// Token vesting contract.
+/// Adapted from Melonport AG <team@melonport.com>
+
+import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+
+contract VestingToken {
+    using SafeMath for uint256;
+
+    ERC20 public token; // Radicle ERC20 contract
+    address public owner; // deployer; can interrupt vesting
```
CodeSandwich, line 13:
```suggestion
    ERC20 public immutable token; // Radicle ERC20 contract
    address public immutable owner; // deployer; can interrupt vesting
```
They never change, no need to keep them in storage

contracts/Governance/VestingToken.sol
```diff
@@ -0,0 +1,119 @@
...
+
+    bool public interrupted; // whether vesting is still possible
+    bool public isVestingStarted; // whether vesting period has begun
+    uint256 public totalVestingAmount; // quantity of vested token in total
+    uint256 public vestingStartTime; // timestamp when vesting is set
+    uint256 public vestingPeriod; // total vesting period in seconds
+    address public beneficiary; // address of the beneficiary
+    uint256 public withdrawn; // quantity of token withdrawn so far
```
CodeSandwich, line 21: If we reversed the logic and stored `unwithdrawn`, we'd elegantly zero the storage on the end of life of the contract. Probably not that important :D

CodeSandwich:
Could you update the changelog as well?

CodeSandwich:
I guess that we do not want to trust the `owner` with holding sufficient number of tokens? If we did, we could skip pushing them through the vesting contract and send them directly from the owner to the beneficiary.

contracts/Governance/VestingToken.sol
```diff
@@ -0,0 +1,119 @@
...
+        require(
+            token.transfer(beneficiary, withdrawable),
+            "VestingToken::withdrawVested: transfer to beneficiary failed"
+        );
+    }
+
+    /// @notice Force withdrawal of vested tokens to beneficiary
+    /// @notice Send remainder back to owner
+    /// @notice Prevent further vesting
+    function terminateVesting() external onlyOwner vestingStarted notInterrupted {
```
CodeSandwich, line 101: The current API won't be very useful for our use case. This function should accept the timestamp of when the termination should be enacted.

contracts/Governance/VestingToken.sol
```diff
@@ -0,0 +1,119 @@
...
+
+    /// @notice Grant vesting tokens to the beneficiary.
+    ///
+    /// @param ofBeneficiary Address of beneficiary
+    /// @param ofTokenAmount Amount of tokens
+    /// @param ofVestingPeriod Vesting period in seconds from vestingStartTime
+    function grantTokens(
+        address ofBeneficiary,
+        uint256 ofTokenAmount,
+        uint256 ofVestingPeriod
```
CodeSandwich, line 70: This function should accept the timestamp of when the vesting period should start.

contracts/Governance/VestingToken.sol
cloudhead: This terminates it immediately, for example in the case an employee leaves the company. We can call it on the date they leave, or do you think that's a problem?

contracts/Governance/VestingToken.sol
cloudhead: Same question here, are you saying it should be back-dated to the vesting start date? The way we were going to do this afaik is to send the already-vested tokens directly, and only use the contract for the unvested ones. But I guess it will be impossible to calculate it exactly in that case?

contracts/Governance/VestingToken.sol
cloudhead: Can you elaborate? Aren't they kept in storage regardless?

contracts/Governance/VestingToken.sol
cloudhead: Good point!

contracts/Governance/VestingToken.sol
CodeSandwich: We know in advance when exactly a person is going to leave the company. The current API doesn't help us with that use case, it requires somebody to actually send a message roughly at that moment. It's imprecise and cumbersome. The smart contract could do that for us automatically and with perfect precision.

contracts/Governance/VestingToken.sol
CodeSandwich: No, the `immutable` variables are baked into the contract as constants when deployed

contracts/Governance/VestingToken.sol
CodeSandwich: Yes, back-dating to the exact moment considered to be the release of the token is one part. There's no need for us send the already-vested tokens directly if we design this API right. The other is that a cliff period may end after the release of the token and with this API somebody will need to track it and try to snipe that moment with a message.

contracts/Governance/VestingToken.sol
cloudhead: 73824c9

contracts/Governance/VestingToken.sol
cloudhead: Will tackle this in a separate PR

contracts/Governance/VestingToken.sol
cloudhead: Done in 73824c9

cloudhead:
> I guess that we do not want to trust the `owner` with holding sufficient number of tokens? If we did, we could skip pushing them through the vesting contract and send them directly from the owner to the beneficiary.

we don't want this indeed, because this would allow the owner to withhold funds from the beneficiary

contracts/Governance/VestingToken.sol
CodeSandwich: Cool! Now almost all the variables can be `immutable` to cut gas usage!

contracts/Governance/VestingToken.sol
cloudhead: Ah cool!

contracts/Governance/VestingToken.sol
cloudhead: Ahahaha maybe something we can do afterwards if we have extra time!

contracts/Governance/VestingToken.sol
cloudhead: contracts/Governance/VestingToken.sol:55:13: TypeError: Immutable variables cannot be read during contract creation time, which means they cannot be read in the constructor or any function or modifier called from it.
--
  | token.transferFrom(msg.sender, address(this), _amount),

CodeSandwich:
> we don't want this indeed, because this would allow the owner to withhold funds from the beneficiary

In case of a bad actor it's still possible with `terminateVesting`. All the power is in the hands of the owner regardless of the actual funds flow. Sending funds from the owner to the receiver directly only on withdrawal would simplify the logic.

contracts/Governance/VestingToken.sol
cloudhead: 03be174

cloudhead:
> > we don't want this indeed, because this would allow the owner to withhold funds from the beneficiary
> 
> In case of a bad actor it's still possible with `terminateVesting`. All the power is in the hands of the owner regardless of the actual funds flow. Sending funds from the owner to the receiver directly only on withdrawal would simplify the logic.

The owner cannot block withdrawal of vested tokens currently. If we keep the funds with the owner, they can block even vested tokens from being withdrawn.

CodeSandwich:
Yes, that's right. But to be honest I'm more afraid of bugs and mistakes than of the owner being a bad actor. Any kind of malicious actions or failure of delivery would end up in the court, no matter how many tokens would be affected. On the other hand sending all vested tokens to the contract reduces room for human error. I think that you're right, that flow complexity is probably worth keeping for that reason alone.

cloudhead:
Yeah, the thing is I don't see the point of using a contract if the owner can decide how many tokens you can take out, then we don't need a contract at all..

contracts/Governance/VestingToken.sol
CodeSandwich: https://github.com/radicle-dev/radicle-contracts/pull/84/commits/e604d1207da0998087d5436a92b7b26d5dd1fcf7 :)

contracts/Governance/VestingToken.sol
```diff
@@ -0,0 +1,115 @@
...
+    /// @param _vestingStartTime Vesting start time in seconds since Epoch
+    constructor(
+        address _token,
+        address _owner,
+        address _beneficiary,
+        uint256 _amount,
+        uint256 _vestingStartTime,
+        uint256 _vestingPeriod
+    ) {
+        require(_vestingStartTime < block.timestamp, "Vesting start time must be in the past");
```
CodeSandwich, line 53:
```suggestion
        require(_vestingStartTime < block.timestamp, "Vesting start time must be in the past");
        require(_vestingPeriod > 0, "Vesting period time must be positive");
```

contracts/Governance/VestingToken.sol
```diff
@@ -0,0 +1,115 @@
...
+    /// @notice Create a vesting allocation of tokens.
+    ///
+    /// @param _token Address of token being vested
+    /// @param _beneficiary Address of beneficiary
+    /// @param _amount Amount of tokens
+    /// @param _vestingPeriod Vesting period in seconds from vestingStartTime
+    /// @param _vestingStartTime Vesting start time in seconds since Epoch
+    constructor(
+        address _token,
+        address _owner,
```
CodeSandwich, line 47: With the current API you can create the vesting contract without becoming an owner. Is it intended? It raises a flag when reading, because then you'll give the vested tokens, but the owner will take them if terminated.

contracts/Governance/VestingToken.sol
```diff
@@ -52,12 +52,13 @@ contract VestingToken {
     ) {
         require(_vestingStartTime < block.timestamp, "Vesting start time must be in the past");
         require(_amount > 0, "VestingToken::constructor: amount must be positive");
+        ERC20 erc20 = ERC20(_token);
         require(
-            token.transferFrom(msg.sender, address(this), _amount),
+            erc20.transferFrom(msg.sender, address(this), _amount),
             "VestingToken::constructor: token deposit failed"
         );
```
cloudhead, line 59: lol clever

contracts/Governance/VestingToken.sol
cloudhead: I think that's fine. This allows for not having an owner too, which is useful, eg. when granting vesting tokens to the community.

cloudhead:
Ok all done from my side.

contracts/Governance/VestingToken.sol
CodeSandwich: Cool trick!

CodeSandwich (approved):

CodeSandwich:
Awesome!

PR #86, opened 2021-01-07
Updated registrar

cloudhead:
This allows the registrar fees to be changed. It also allows payment in Radicle.


CodeSandwich (approved):
Looking good!

PR #91, opened 2021-01-11
Make Registrar exchange, admin, oracle upgradable

cloudhead:
The Registrar contract itself won't be upgradable, but its parts will be!

### DIFF
--- a/contracts/Registrar.sol
+++ b/contracts/Registrar.sol
@@ -20,8 +20,14 @@ contract Registrar {
     /// The Rad ERC20 token.
     Rad public immutable rad;
 
-    /// The namehash of the domain this registrar owns(eg. radicle.eth).
-    bytes32 public immutable rootNode;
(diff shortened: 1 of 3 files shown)


CodeSandwich (commented):
Looks good, but should the registrar be non-upgradable? If so, shouldn't it get at least discardable and allow changing the owner of the ENS subdomain?

cloudhead:
Hmm, I wonder how that would work?

CodeSandwich:
Something as simple as this would allow us to create a new version of the registrar and make the ENS acknowledge it:
```
function setDomainOwner(address newOwner) public adminOnly {
    ens.setSubnodeOwner(keccak256("eth"), keccak256("radicle"), newOwner);
}
```

cloudhead:
Done!

CodeSandwich (approved):
Nice!

PR #92, opened 2021-01-12
A price oracle that always fails

cloudhead:
This oracle can be used in the registrar to effectively disable registration via ETH.

### DIFF
--- a/contracts/FailingOracle.sol
+++ b/contracts/FailingOracle.sol
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: GPL-3.0-only
 // solhint-disable no-empty-blocks
+// solhint-disable no-unused-vars
 pragma solidity ^0.7.5;
 
 import "./PriceOracle.sol";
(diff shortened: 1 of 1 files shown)


contracts/FailingOracle.sol
```diff
@@ -0,0 +1,15 @@
...
+
+import "./PriceOracle.sol";
+
+contract FailingOracle is PriceOracle {
+    constructor() {}
+
+    function updatePrices() public override {}
+
+    function consultUsdEth(uint256 _usdAmount) public view override returns (uint256) {
+        revert("FailingOracle::consultUsdEth: this oracle always fails");
```
CodeSandwich, line 13: This message is correct, but useless and confusing for the the end user. We're using this oracle in one place anyway and for one purpose, how about:
```suggestion
        revert("ETH not supported yet, please use RAD");
```

CodeSandwich (approved):

PR #94, opened 2021-01-13
Update Registrar to only take Radicle tokens

cloudhead:


contracts/Registrar.sol
```diff
@@ -48,65 +46,47 @@ contract Registrar {
...
+        ERC20Burnable rad = ERC20Burnable(radAddress);
         uint256 fee = registrationFeeRad;
 
         require(rad.balanceOf(msg.sender) >= fee, "Transaction includes registration fee");
         require(valid(name), "Name must be valid");
 
         _register(keccak256(bytes(name)), owner);
 
-        rad.burn(fee);
+        rad.burnFrom(msg.sender, fee);
```
cloudhead, line 83: I think this was a bug!

contracts/Rad.sol
```diff
@@ -13,9 +13,4 @@ contract Rad is ERC20 {
     constructor(address account, uint256 initialSupply) ERC20("Radicle", "RADICLE") {
         _mint(account, initialSupply.mul(10**uint256(decimals())));
     }
-
-    /// Burn the given amount of tokens from the sender account.
-    function burn(uint256 amount) public {
-        _burn(msg.sender, amount);
-    }
 }
```
cloudhead, line 16: Will remove this contract later, but for now we need it for the tests and deploy script.

contracts/Registrar.sol
```diff
@@ -48,65 +46,47 @@ contract Registrar {
...
         baseLabel = _baseLabel;
         parentNode = _parentNode;
         domain = namehash(_parentNode, _baseLabel);
         admin = adminAddress;
     }
 
     function initialize() public {
-        oracle.updatePrices();
+        // This can be used in the future to initialize the exchange and oracle.
     }
```
CodeSandwich, line 66: How is this different from adding things in the constructor?

contracts/Registrar.sol
```diff
@@ -1,24 +1,22 @@
...
-    PriceOracle public oracle;
+    address public oracleAddress;
 
     /// The Rad/Eth exchange.
-    Exchange public exchange;
+    address public exchangeAddress;
 
     /// The Rad ERC20 token.
-    Rad public immutable rad;
+    address public immutable radAddress;
```
CodeSandwich, line 19: I think that it can be of type `Rad`, AFAIK the contract type is just an address, so they'll compile to the same ABI

contracts/Registrar.sol
cloudhead: Good question, aren't there limitations to what you can do in the constructor? I'm trying to remember why I did it this way :thinking:

contracts/Registrar.sol
cloudhead: Oh so internally, the storage layout is the same? That makes sense..

contracts/Registrar.sol
cloudhead: Perhaps one thing to think about is: will we have to have a function for when a new oracle is set, to initialize it? Eg. if I do `registrar.setPriceOracle(addr)`, I may then need to do `registrar.initializePriceOracle()` or something :thinking: 

Although we call `updatePrices` in `register`, so maybe that's not necessary?

contracts/Registrar.sol
CodeSandwich: I've double checked it and it's correct, contracts are just addresses: https://docs.soliditylang.org/en/v0.7.5/abi-spec.html#mapping-solidity-to-abi-types

contracts/Registrar.sol
CodeSandwich: IMO the right solution is to call `oracle.updatePrices();` inside `registrar.setPriceOracle(addr)`, otherwise we're creating a whole class of edge cases to exploit. That's why `initialize()` looks so suspicious in the first place.

cloudhead: We'll have to do that when we upgrade the contract then, because we're just working with addresses atm.

contracts/Registrar.sol
```diff
@@ -15,8 +15,8 @@ contract Registrar {
     /// The Rad/Eth exchange.
     address public exchangeAddress;
 
-    /// The Rad ERC20 token.
-    address public immutable radAddress;
+    /// The Radicle ERC20 token.
+    ERC20Burnable public immutable rad;
```
CodeSandwich, line 19: That one is hairy, RadicleToken isn't really `ERC20Burnable`, it lacks many non-ERC20 functions present in OpenZeppelin's implementation. It will work thanks to duck typing of the dynamic dispatch and because we simply don't use the unimplemented functions, but it's VERY ugly and misleading. The best solution would be to create `IERC20Burnable` deriving from `IERC20`.

BTW I've noticed that RadicleToken IS NOT an ERC-20 token because e.g. it exposes `DECIMALS()` instead of `decimals()`. I haven't fixed it yet, but we absolutely have to straighten it before any release. Having a proper interface would catch that.

cloudhead: For now, since we're using `Rad.sol`, it's okay. We can change it when we migrate to `RadicleToken.sol`, and just have it as `RadicleToken`.

`RadicleToken` definitely is ERC20 compatible, I think `decimals()` is optional. We should however maybe enforce it via the interface.

CodeSandwich: That's fair, for now let's keep it in mind.

> RadicleToken definitely is ERC20 compatible, I think decimals() is optional. We should however maybe enforce it via the interface.

Nice, I wasn't aware that parts of ERC-20 are optional. But we're still missing `totalSupply` :P

cloudhead: Yeah just noticed :joy: 

We should probably implement it, although it doesn't seem to pose a problem for COMP, I'm guessing because wallets don't actually ever call that function..

CodeSandwich (approved):

PR #105, opened 2021-01-15
Release script

CodeSandwich:


src/deploy-to-network.ts
```diff
@@ -102,3 +128,31 @@ export async function phase0(): Promise<void> {
...
+  const token = await askForAddress("of the token");
+  const owner = await askForAddress("of the owner of all created vestings");
+  do {
+    const beneficiary = await askForAddress("of beneficiary");
+    const amount = await askForBigNumber("amount to vest");
+    const vestingStartTime = await askForBigNumber(
+      "vesting start time in seconds since epoch"
+    );
+    const vestingPeriod = await askForBigNumber("vesting period in seconds");
+    const cliffPeriod = await askForBigNumber("cliff period in seconds");
```
cloudhead, line 145: Since everyone has the same cliff and vesting period, maybe we can have these as constants in the deploy script?

CodeSandwich: Fixed

src/deploy-to-network.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  const ens = ENS__factory.connect(ensAddr, signer);
+  const registrar = await deploy.deployRegistrar(
+    signer,
+    constants.AddressZero, // oracle not used yet
+    constants.AddressZero, // exchange not used yet
+    token.address,
+    ens,
+    "eth",
+    "radicle",
```
cloudhead, line 124: Don't these have to be a namehash?

src/deploy-to-network.ts
```diff
@@ -0,0 +1,162 @@
...
+    govAddr
+  );
+  printDeployed("registrar", registrar);
+
+  await provider.close();
+}
+
+export async function vestedTokens(): Promise<void> {
```
cloudhead, line 132:
```suggestion
export async function vestingTokens(): Promise<void> {
```
The yarn task should be renamed too.

src/deploy-to-network.ts
```diff
@@ -0,0 +1,162 @@
...
+    await deploy.deployVestingToken(
+      signer,
+      token,
+      owner,
+      beneficiary,
+      amount,
+      vestingStartTime,
+      vestingPeriod,
+      cliffPeriod
+    );
```
cloudhead, line 159: We should print the address here of each vesting contract

src/deploy-to-network.ts
cloudhead: Nevermind, it's done in the function.

cloudhead: Nevermind, it's done in the function.

src/deploy.ts
```diff
@@ -49,77 +49,114 @@ export interface DeployedContracts {
...
+    ensUtils.nameHash(label + "." + domain),
     oracle,
-    exchange.address,
-    rad,
-    signerAddr
+    exchange,
+    token,
+    admin
   );
+  await setSubnodeOwner(ens, domain, label, registrar.address);
```
cloudhead, line 126: This will only work if we deploy using the owner of `radicle.eth` :thinking: 

Maybe this line should be removed, and only be used in the test-ens scenario? Because in the real deployment, we will probably manually transfer ownership, or perhaps it could be done in a separate deployment step where the owner can connect their wallet?

CodeSandwich: I've fixed it, but haven't pushed the changes, please take a look again

src/deploy-to-network.ts
CodeSandwich: Fixed

src/deploy-to-network.ts
CodeSandwich: Fixed

src/deploy-to-network.ts
CodeSandwich: That's right, `deployRegistrar` does that

contracts/Registrar.sol
```diff
@@ -147,7 +138,7 @@ contract Registrar {
 
     /// Set the owner of the domain.
     function setDomainOwner(address newOwner) public adminOnly {
-        ens.setOwner(namehash(parentNode, baseLabel), newOwner);
+        ens.setOwner(domain, newOwner);
```
cloudhead, line 141: Should we be using `ens.transferFrom` actually? :thinking:

src/deploy-to-network.ts
```diff
@@ -0,0 +1,162 @@
...
+  );
+  const cliffPeriod = await askForBigNumber(
+    "cliff period of all created vestings in seconds"
+  );
+  do {
+    const beneficiary = await askForAddress("of beneficiary");
+    const amount = await askForBigNumber("amount to vest");
+    const vestingStartTime = await askForBigNumber(
+      "vesting start time in seconds since epoch"
+    );
```
cloudhead, line 166: Can we have a confirmation here, which shows the beneficiary + amount etc.?

src/deploy.ts
cloudhead: Looks good.

contracts/Registrar.sol
CodeSandwich: I'm sure that `setOwner` is used here correctly. The registrar will be the owner of the `radicle.eth` domain and it's the only address entitled to pass it to someone else.

What does `transferFrom` do? The ENS docs only mention it once, in context of the ERC-721 tokens API.

src/deploy-to-network.ts
CodeSandwich: Fixed

contracts/Registrar.sol
cloudhead: I asked one of the ENS people, and the answer was "`transferFrom` should be enough". :shrug:

src/deploy-to-network.ts
```diff
@@ -0,0 +1,198 @@
...
+  console.log("Registrar owns the ", domain, "domain");
+
+  await provider.close();
+}
+
+export async function vestingTokens(): Promise<void> {
+  console.log("The wallet owner will be the one providing tokens for vesting");
+  const provider = await walletConnectProvider();
+  const signer = provider.getSigner();
+  const tokenAddr = await askForAddress("of the token");
```
cloudhead, line 165:
```suggestion
  const tokenAddr = await askForAddress("of the Radicle token contract");
```

src/deploy-to-network.ts
```diff
@@ -0,0 +1,198 @@
...
+  const token = ERC20__factory.connect(tokenAddr, signer);
+  const owner = await askForAddress("of the owner of all created vestings");
```
cloudhead, line 167:
```suggestion
  const owner = await askForAddress("of the vesting contract(s) admin");
```

src/deploy-to-network.ts
```diff
@@ -0,0 +1,198 @@
...
+        tokenAddr,
+        owner,
+        beneficiary,
+        amount,
+        vestingStartTime,
+        vestingPeriod,
+        cliffPeriod
+      )
+    );
+    console.log(beneficiary, "has", amount.toString(), "tokens vested");
```
cloudhead, line 195:
```suggestion
    console.log(beneficiary, "has", amount.toString(), "tokens vesting");
```

contracts/Registrar.sol
CodeSandwich: Fixed

PR #109, opened 2021-02-02
Add radicle token tests

CodeSandwich:


cloudhead (approved):
Awesome. Are these from compound?

CodeSandwich:
Yes, these scenarios are taken from compund, they've created a good set

PR #116, opened 2021-02-05
Create factory contract for phase0

CodeSandwich:
Due to what I believe is a hardhat bug (https://github.com/nomiclabs/hardhat/issues/1227) a few tests aren't verifying transaction rejection messages anymore.

### DIFF
--- a/contracts/deploy/phase0.sol
+++ b/contracts/deploy/phase0.sol
@@ -1,0 +1,69 @@
+// spdx-license-identifier: gpl-3.0-only
+pragma solidity ^0.7.5;
+
+import {ENS} from "@ensdomains/ens/contracts/ENS.sol";
+import {ERC20Burnable} from "@openzeppelin/contracts/token/ERC20/ERC20Burnable.sol";
+import {Governor} from "../Governance/Governor.sol";
+import {RadicleToken} from "../Governance/RadicleToken.sol";
+import {Timelock} from "../Governance/Timelock.sol";
+import {Treasury} from "../Governance/Treasury.sol";
+import {Registrar} from "../Registrar.sol";
+
+contract Phase0 {
+    RadicleToken public immutable token;
+    Timelock public immutable timelock;
+    Governor public immutable governor;
+    Registrar public immutable registrar;
+
+    address public immutable tokensHolder;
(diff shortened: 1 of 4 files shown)


contracts/deploy/phase0.sol
```diff
@@ -0,0 +1,72 @@
...
+        );
+        uint8 governorNonce = 3;
+        bytes memory govAddrPayload = abi.encodePacked(hex"d694", address(this), governorNonce);
+        address govAddr = address(uint256(keccak256(govAddrPayload)));
+
+        RadicleToken _token = new RadicleToken(_tokensHolder);
+        Timelock _timelock = new Timelock(govAddr, _timelockDelay);
+        Governor _governor = new Governor(address(_timelock), address(_token), _governorGuardian);
+        require(address(_governor) == govAddr, "Governor deployed under an unexpected address");
+        Treasury _treasury = new Treasury(address(_timelock));
```
cloudhead, line 47: Can we actually use the Timelock already here?

CodeSandwich: Fixed, just removed treasury deployment, we don't need anything else done

contracts/deploy/phase0.sol
```diff
@@ -13,7 +13,6 @@ contract Phase0 {
     RadicleToken public immutable token;
```
cloudhead, line 13: Code is correct, but commit message isn't ;)

CodeSandwich: What do you mean?

cloudhead: It says "Remove timelock deployment"

CodeSandwich: Yes and the commit with that message does exactly that :thinking: Anyway we shouldn't sweat too much over any commit messages except the first one, they will be squashed before merging.

cloudhead: The commit removes the treasury, not the timelock :grimacing:

cloudhead (approved):

contracts/deploy/phase0.sol
CodeSandwich: Right :facepalm:

PR #120, opened 2021-02-10
Run prettier

cloudhead:


cloudhead:
Adding prettier as part of our toolchain was such a bad idea..

CodeSandwich (approved):
Nah, non-negotiable code formatters are a huge relief

contracts/Governance/RadicleToken.sol
```diff
@@ -122,8 +122,8 @@ contract RadicleToken {
         return DECIMALS;
     }
 
-    /* @notice DOMAIN_SEPARATOR */
-    function DOMAIN_SEPARATOR() public view returns (bytes32) {
+    /* @notice domainSeparator */
+    function domainSeparator() public view returns (bytes32) {
```
cloudhead, line 126: I think this is like, a convention / part of the API, so we should keep it in uppercase: https://github.com/ethereum/EIPs/blob/master/assets/eip-712/Example.sol#L35

CodeSandwich: But on the other hand ERC-20's `decimals` always returns a constant, but is lowercase. Maybe the SCREAMING_SNAKE_CASE convention may not be best suited for the public parts of the API, there are no public fields anyway, just functions :thinking:

Does it need to be public? If it doesn't need to, then we don't need to bend to any conventions for future proofing.

cloudhead: hmm maybe just drop a message on the radicle token channel

cloudhead: I'd keep it public just in case, but maybe the case doesn't matter.

cloudhead: in the EIP: https://eips.ethereum.org/EIPS/eip-712#definition-of-domainseparator it's actually camelCase..

MrChico: In https://eips.ethereum.org/EIPS/eip-2612#specification it's SCREAMING_SNAKE_CASE

PR #121, opened 2021-02-10
Optional improvements from audit

cloudhead:
* Couldn't use `calldata` everywhere, as it says "callstack too deep"  on `propose`.
* I actually changed the threshold logic, because it doesn't match the other threshold check. So I think the error message was correct, and the logic was wrong. See: https://github.com/radicle-dev/radicle-contracts/blob/master/contracts/Governance/Governor.sol#L299 -- without this change, if you have exactly the threshold, you can neither propose nor cancel..

### DIFF
--- a/contracts/Governance/Governor.sol
+++ b/contracts/Governance/Governor.sol
@@ -71,8 +71,6 @@ contract Governor {
 
     /// @notice Change proposal
     struct Proposal {
-        // Unique id for looking up a proposal
-        uint256 id;
         // Creator of the proposal
         address proposer;
         // The timestamp that the proposal will be available for execution, set once the vote succeeds
@@ -201,7 +199,8 @@ contract Governor {
         proposalCount++;
         Proposal storage newProposal = proposals[proposalCount];
 
-        newProposal.id = proposalCount;
+        uint256 proposalId = proposalCount;
+
         newProposal.proposer = msg.sender;
         newProposal.eta = 0;
         newProposal.targets = targets;
@@ -215,10 +214,10 @@ contract Governor {
         newProposal.canceled = false;
         newProposal.executed = false;
 
-        latestProposalIds[newProposal.proposer] = newProposal.id;
+        latestProposalIds[newProposal.proposer] = proposalId;
 
         emit ProposalCreated(
-            newProposal.id,
+            proposalId,
             msg.sender,
             targets,
             values,
@@ -228,7 +227,7 @@ contract Governor {
             endBlock,
             description
         );
-        return newProposal.id;
+        return proposalId;
     }
 
     function queue(uint256 proposalId) public {
--- a/contracts/Governance/Governor.sol
+++ b/contracts/Governance/Governor.sol
@@ -295,9 +295,9 @@ contract Governor {
         Proposal storage proposal = proposals[proposalId];
         require(
             msg.sender == guardian ||
-            // Allows anyone to cancel a proposal if the voting power of the
-            // proposer dropped below the threshold after the proposal was
-            // submitted.
+                // Allows anyone to cancel a proposal if the voting power of the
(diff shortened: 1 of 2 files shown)


CodeSandwich (approved):
These are good improvements :+1:

contracts/Governance/Governor.sol
```diff
@@ -170,7 +168,7 @@ contract Governor {
         string memory description
     ) public returns (uint256) {
         require(
-            token.getPriorVotes(msg.sender, sub256(block.number, 1)) > proposalThreshold(),
+            token.getPriorVotes(msg.sender, sub256(block.number, 1)) >= proposalThreshold(),
```
cloudhead, line 171: Notice this logic change: `>` changed to `>=`. Please double check this makes sense.

contracts/Governance/Governor.sol
```diff
@@ -288,8 +288,8 @@ contract Governor {
     function cancel(uint256 proposalId) public {
         ProposalState _state = state(proposalId);
         require(
-            _state != ProposalState.Executed,
-            "Governor::cancel: cannot cancel executed proposal"
+            _state == ProposalState.Queued,
+            "Governor::cancel: cannot cancel proposal that isn't queued"
         );
```
cloudhead, line 293: This change comes from the trail of bits audit.

contracts/Governance/Governor.sol
CodeSandwich: It does, `>=` = can propose, `<` = can't propose, existing proposals are moot

contracts/Governance/Governor.sol
CodeSandwich: It doesn't make sense to me, why shouldn't the guardian be able to shoot down proposals in other states?

CodeSandwich: No, I actually don't see it. If the proposer loses power to push forward changes, why do we need to wait for their proposal to reach this late state to acknowledge it?

cloudhead: So the thing is, if you check the Timelock cancelTransaction code and queueTransaction, if we call cancel before it's queued, it's still possible to queue, and then that will just "undo" the cancel, in the timelock.

CodeSandwich: I agree that canceling something not queued on the timelock is a noop. But cancelling also sets the state of the proposal on the governor side. For example cancelling a proposal, which is active should prevent it from being able to be voted on and proceeded.

CodeSandwich (approved):

contracts/Governance/Governor.sol
```diff
@@ -296,22 +295,27 @@ contract Governor {
...
-            timelock.cancelTransaction(
-                proposal.targets[i],
-                proposal.values[i],
-                proposal.signatures[i],
-                proposal.calldatas[i],
-                proposal.eta
-            );
-        }
 
+        if (_state == ProposalState.Queued) {
```
MrChico, line 308: Note that this is a semantic change. In the current version, you can cancel proposals that are `Queued` or `Expired`, whereas with this change you can only cancel `Queued` transactions. Of course, `Expired` proposals can't be executed anyway, so canceling them is mainly an cosmetic operation (unless the admin of the Timelock is changed).

I'm not sure how I feel about this change tbh. I don't think the gas efficiency concern brought up by ToB is particularly relevant. We'll remove the mention of this in our report.

cloudhead: Hmm, I'm fine with reverting the change. I thought that since `proposal.canceled = true` is still run, this wouldn't actually change the semantics. :eyes:

MrChico: Since proposals aren't marked as cancelled in the `timelock`, it would only make a difference when upgrading the timelock admin

PR #122, opened 2021-02-11
Optimize gas in phase0 contracts

CodeSandwich:
There's quite a lot going on in this PR:
- Switch all timestamp and block numbers to uint64
- Switch all radicle amounts to uint96
- This required some expansion of the `SafeMath.sol` file, the upside is that the contracts don't need to reimplement the safe operations anymore
- Some events have fixed IMO lacking `indexed` fields
- I needed to fix the order of items in contracts, it was impossible to follow the order of storage fields or find events
- Finally, reordered storage fields to better fill slots and group variables, which are accessed at the same time
- Coincidentally it fixed some revocation message tests ¯\\\_(ツ)_/¯

### DIFF
--- a/contracts/Governance/Governor.sol
+++ b/contracts/Governance/Governor.sol
@@ -28,34 +28,22 @@
 pragma solidity ^0.7.5;
 pragma experimental ABIEncoderV2;
 
+import "../libraries/SafeMath.sol";
+
 contract Governor {
+    using SafeMath for uint256;
+    using SafeMath96 for uint96;
+    using SafeMath64 for uint64;
+
     /// @notice The name of this contract
     string public constant NAME = "Radicle Governor";
 
-    /// @notice The number of votes in support of a proposal required in order for a quorum to be reached and for a vote to succeed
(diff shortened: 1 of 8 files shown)


cloudhead:
I thought `uint96` was more expensive than `uint256` due to having to fill the rest of the slot with zeros? :thinking:

CodeSandwich:
It's complicated. Conversion between uint96 and uint256 on the stack is a noop. Having a uint96 in a densely packed storage slot has a cost, because it must be inserted or extracted with binary operations. But this is few orders of magnitude cheaper than using an additional slot.

cloudhead:
I'll recap my notes here from the chat:

* I'm not comfortable merging so many changes at this point in time, unless critical
* There is some code that is moved around that we can avoid moving to make the diff smaller
* I'd be interested to see how much gas we're saving here
* Requiring three versions of SafeMath is overkill imo
* I'd prefer to focus the changes to better packing the Proposal struct, and making any other gas optimizations post-launch

PR #125, opened 2021-02-12
Clean up license headers

CodeSandwich:
As proposed by @cloudhead
CC @angedupre


cloudhead (approved):

CodeSandwich:
BTW we have a copy of the Uniswap code explicitly commented as such. It's originally licensed under GPL-3, should we slap the MIT SPDX header anyway? Or should we leave it headerless and just keep the comment about its origin?

cloudhead:
Hmm maybe that's a problem :thinking:
Should we just use GPLv3 everywhere?

CodeSandwich:
Switched all to GPL-3, I hope that it's all legal now :laughing:

PR #127, opened 2021-02-15
Make certain allocations at genesis

cloudhead:


CodeSandwich (approved):
Looking good, I'm glad that it's enough :)

PR #129, opened 2021-03-05
Add Link ID claims contract

CodeSandwich:
Implementation of a smart contract for https://github.com/radicle-dev/radicle-link/pull/557

### DIFF
--- a/contracts/Claims.sol
+++ b/contracts/Claims.sol
@@ -1,0 +1,27 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.7.5;
+
+// @notice Used to create Radicle Link identity claims
+contract Claims {
+    /// @notice Emitted on every Radicle Link identity claim
+    /// @param addr The account address which made the claim
+    event Claimed(address indexed addr);
+
+    /// @notice Creates a new claim of a Radicle Link identity.
+    /// Every new claim invalidates previous ones made with the same account.
+    /// The claims have no expiration date and don't need to be renewed.
+    /// If either `format` is unsupported or `payload` is malformed as per `format`,
+    /// the previous claim is revoked, but a new one isn't created.
+    /// Don't send a malformed transactions on purpose, to properly revoke a claim see `format`.
+    /// @param format The format of `payload`, currently supported values:
+    /// - `1` - `payload` is exactly 20 bytes and contains an SHA-1 Radicle Identity root hash
+    /// - `2` - `payload` is exactly 32 bytes and contains an SHA-256 Radicle Identity root hash
+    /// To revoke a claim without creating a new one, pass payload `0`,
+    /// which is guaranteed to not match any existing identity.
+    /// @param payload The claim payload
+    function claim(uint256 format, bytes calldata payload) public {
+        format;
+        payload;
+        emit Claimed(msg.sender);
+    }
+}
--- a/src/deploy.ts
+++ b/src/deploy.ts
@@ -9,6 +9,7 @@ import {
   ContractReceipt,
   Signer,
 } from "ethers";
+import { Claims } from "../contract-bindings/ethers/Claims";
 import { ENS } from "../contract-bindings/ethers/ENS";
 import { EthPool } from "../contract-bindings/ethers/EthPool";
 import { Exchange } from "../contract-bindings/ethers/Exchange";
@@ -20,6 +21,7 @@ import { Timelock } from "../contract-bindings/ethers/Timelock";
 import { VestingToken } from "../contract-bindings/ethers/VestingToken";
 import {
   BaseRegistrarImplementation__factory,
+  Claims__factory,
   ENSRegistry__factory,
   Erc20Pool__factory,
   Erc20Pool,
@@ -60,6 +62,7 @@ export interface DeployedContracts {
   ens: ENS;
   ethPool: EthPool;
   erc20Pool: Erc20Pool;
+  claims: Claims;
 }
 
 export async function deployAll(signer: Signer): Promise<DeployedContracts> {
@@ -82,8 +85,9 @@ export async function deployAll(signer: Signer): Promise<DeployedContracts> {
   await transferEthDomain(ens, label, registrar.address);
   const ethPool = await deployEthPool(signer, 10);
   const erc20Pool = await deployErc20Pool(signer, 10, rad.address);
+  const claims = await deployClaims(signer);
 
-  return { gov, rad, exchange, registrar, ens, ethPool, erc20Pool };
+  return { gov, rad, exchange, registrar, ens, ethPool, erc20Pool, claims };
 }
 
 export async function deployRadicleToken(signer: Signer, account: string): Promise<RadicleToken> {
@@ -283,6 +287,10 @@ export async function deployPhase0(
       { gasLimit: 10 * 10 ** 6 }
     )
   );
+}
+
+export async function deployClaims(signer: Signer): Promise<Claims> {
+  return deployOk(new Claims__factory(signer).deploy());
 }
 
 async function deployOk<T extends Contract>(contractPromise: Promise<T>): Promise<T> {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ with regard to its JavaScript and TypeScript bindings' APIs and the Ethereum ABI
 
 ### Added
 - Burning capability to Radicle Token
+- Radicle Link ID claims contract
 
 ### Changed
 - Replaced vesting contract with one adapted from Melonport


NunoAlexandre (commented):
Looking good!
 
I suggest we refer to Link as Radicle Link, it will be helpful to newcomers

contracts/Claims.sol
```diff
@@ -0,0 +1,24 @@
...
+    /// @param version The version of `id` encoding, currently supported values:
+    /// - `1` - an `id` is a SHA-1 root hash. The excess high bytes are zeros, e.g. for hash
+    /// `fb3102b74d7254eed7f18a31a3ba1ea946bb1a99` the passed `id` is
+    /// `000000000000000000000000fb3102b74d7254eed7f18a31a3ba1ea946bb1a99`
+    /// - `2` - an `id` is a SHA-256 root hash
+    /// @param id The claimed Link identity. To revoke a claim without creating a new one, pass `0`,
+    function claim(uint8 version, uint256 id) public {
+        version;
+        id;
+        emit Claimed(msg.sender);
```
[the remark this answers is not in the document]
CodeSandwich: We need to fetch the transaction, which caused this event to be emitted and look at its parameters. The events aren't signed, so we can't trust their content anyway.

src/deploy.ts
```diff
@@ -285,6 +287,10 @@ export async function deployPhase0(
   );
 }
 
+export async function deployClaims(signer: Signer): Promise<Claims> {
+  return deployOk(new Claims__factory(signer).deploy());
+}
```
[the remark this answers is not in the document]
CodeSandwich: Fixed

CodeSandwich:
> I suggest we refer to Link as Radicle Link, it will be helpful to newcomers

Fixed

contracts/Claims.sol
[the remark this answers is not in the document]
NunoAlexandre: I see! thanks :)

NunoAlexandre (approved):
nice!

contracts/Claims.sol
```diff
@@ -0,0 +1,26 @@
...
+    /// @notice Creates a new claim of a Radicle Link identity.
+    /// Every new claim invalidates previous ones made with the same account.
+    /// The claims have no expiration date and don't need to be renewed.
+    /// @param version The version of `root` encoding, currently supported values:
+    /// - `1` - `root` is a SHA-1 root hash. The excess high bytes are zeros, e.g. for hash
+    /// `fb3102b74d7254eed7f18a31a3ba1ea946bb1a99` the passed `root` is
+    /// `000000000000000000000000fb3102b74d7254eed7f18a31a3ba1ea946bb1a99`
+    /// - `2` - `root` is a SHA-256 root hash
+    /// @param root The claimed Radicle Link identity root.
+    /// To only revoke a claim without creating a new one, use root `0`,
```
NunoAlexandre, line 19:
```suggestion
    /// To revoke a claim without creating a new one, pass root `0`,
```

NunoAlexandre (commented):
Just a small suggestion, ready to go for me otherwise :rocket:

contracts/Claims.sol
CodeSandwich: Fixed

NunoAlexandre (approved):
:star:

contracts/Claims.sol
```diff
@@ -0,0 +1,26 @@
...
+    /// The claims have no expiration date and don't need to be renewed.
+    /// @param version The version of `root` encoding, currently supported values:
+    /// - `1` - `root` is a SHA-1 root hash. The excess high bytes are zeros, e.g. for hash
+    /// `fb3102b74d7254eed7f18a31a3ba1ea946bb1a99` the passed `root` is
+    /// `000000000000000000000000fb3102b74d7254eed7f18a31a3ba1ea946bb1a99`
+    /// - `2` - `root` is a SHA-256 root hash
+    /// @param root The claimed Radicle Link identity root.
+    /// To revoke a claim without creating a new one, pass root `0`,
+    /// which is guaranteed to not match any existing identity.
+    function claim(uint8 version, uint256 root) public {
```
cloudhead, line 21: For future compatibility, do we want to add a `byte32` as the third parameter? This would allow additional information to be anchored potentially.

cloudhead: I would also swap the parameter order, as the `root` is the thing being claimed.

contracts/Claims.sol
```diff
@@ -0,0 +1,26 @@
...
+    /// @notice Creates a new claim of a Radicle Link identity.
+    /// Every new claim invalidates previous ones made with the same account.
+    /// The claims have no expiration date and don't need to be renewed.
+    /// @param version The version of `root` encoding, currently supported values:
+    /// - `1` - `root` is a SHA-1 root hash. The excess high bytes are zeros, e.g. for hash
+    /// `fb3102b74d7254eed7f18a31a3ba1ea946bb1a99` the passed `root` is
+    /// `000000000000000000000000fb3102b74d7254eed7f18a31a3ba1ea946bb1a99`
+    /// - `2` - `root` is a SHA-256 root hash
+    /// @param root The claimed Radicle Link identity root.
+    /// To revoke a claim without creating a new one, pass root `0`,
```
cloudhead, line 19: Can't you pass `uint256(0)` to revoke the claim? Since there's only one valid claim per address?

CodeSandwich: Yes, isn't that exactly what the documentation tells you to do to revoke the claim?

contracts/Claims.sol
CodeSandwich: > do we want to add a byte32 as the third parameter?

That's a good point. You've made me reconsidered it and now I think that we should go one step further and converge into this:

```
function claim(uint8 version, bytes payload) public {
```

- For `version == 1` `payload` must be 20 bytes long and hold the SHA-1 root hash.
- For `version == 2` `payload` must be 32 bytes long and hold the SHA-256 root hash.
- And so on

It's provides extreme elasticity without making the current usage any more complex. The gas overhead is minuscule, it's only extra 32 bytes in the transaction data. What do you think @cloudhead @NunoAlexandre ?

NunoAlexandre: Sounds good to me!

CodeSandwich: Added. Also switched `version` from `uint8` to `uint256`, every calldata parameter takes multiples of 32 bytes anyway.

contracts/Claims.sol
```diff
@@ -0,0 +1,29 @@
...
+    /// @notice Creates a new claim of a Radicle Link identity.
+    /// Every new claim invalidates previous ones made with the same account.
+    /// The claims have no expiration date and don't need to be renewed.
+    /// If either `version` is unsupported or `payload` is malformed as per `version`,
+    /// the previous claim is revoked, but a new one isn't created.
+    /// Don't send a malformed transactions on purpose, to properly revoke a claim see `version`.
+    /// @param version The version of `payload` encoding, currently supported values:
+    /// - `1` - `payload` is exactly 20 bytes and contains an SHA-1 root hash.
+    /// To only revoke a claim without creating a new one, pass root hash `0`,
+    /// which is guaranteed to not match any existing identity.
```
NunoAlexandre, line 19: This block is now repeated below and it should now refer to `payload` instead.

CodeSandwich: Fixed

contracts/Claims.sol
```diff
@@ -0,0 +1,27 @@
...
+    /// The claims have no expiration date and don't need to be renewed.
+    /// If either `version` is unsupported or `payload` is malformed as per `version`,
+    /// the previous claim is revoked, but a new one isn't created.
+    /// Don't send a malformed transactions on purpose, to properly revoke a claim see `version`.
+    /// @param version The version of `payload` encoding, currently supported values:
+    /// - `1` - `payload` is exactly 20 bytes and contains an SHA-1 root hash.
+    /// - `2` - `payload` is exactly 32 bytes and contains an SHA-256 root hash.
+    /// To only revoke a claim without creating a new one, pass root hash `0`,
+    /// which is guaranteed to not match any existing identity.
+    /// @param payload The claim payload
```
NunoAlexandre, line 21: I find that 'version' is quite confusing. What do you think about renaming it to `enconding` or `format`, which can either be `"sha-1"` or `"sha-256"`. That would reduce indirection both in the naming and in the transcription of this argument.

contracts/Claims.sol
cloudhead: Yup, I like this!

contracts/Claims.sol
cloudhead: :confused: I swear I read "Pass 0 as the version"

contracts/Claims.sol
cloudhead: Yeah I was about to say the same, it's more of a "type" or "format" than just version, since it basically describes the content of payload now.

contracts/Claims.sol
```diff
@@ -0,0 +1,27 @@
...
+    /// @notice Emitted on every Radicle Link identity claim
+    /// @param addr The account address which made the claim
+    event Claimed(address indexed addr);
+
+    /// @notice Creates a new claim of a Radicle Link identity.
+    /// Every new claim invalidates previous ones made with the same account.
+    /// The claims have no expiration date and don't need to be renewed.
+    /// If either `version` is unsupported or `payload` is malformed as per `version`,
+    /// the previous claim is revoked, but a new one isn't created.
+    /// Don't send a malformed transactions on purpose, to properly revoke a claim see `version`.
```
cloudhead, line 15: This should say "see root"?

contracts/Claims.sol
cloudhead: I wouldn't use a `string` type however, I think `uint` is fine..

CodeSandwich: I like `format`, fixed.

I too feel like `uint` is better here. It's a little bit cheaper and closer to the EVM metal and a whole class of text formatting and encoding bugs just goes away because integers are unambiguous.

contracts/Claims.sol
CodeSandwich: It's a reference to a parameter, now it says ``see `format` ``, is it ok?

cloudhead (approved):
:shipit:

contracts/Claims.sol
```diff
@@ -0,0 +1,27 @@
...
+    /// @notice Creates a new claim of a Radicle Link identity.
+    /// Every new claim invalidates previous ones made with the same account.
+    /// The claims have no expiration date and don't need to be renewed.
+    /// If either `format` is unsupported or `payload` is malformed as per `format`,
+    /// the previous claim is revoked, but a new one isn't created.
+    /// Don't send a malformed transactions on purpose, to properly revoke a claim see `format`.
+    /// @param format The format of `payload`, currently supported values:
+    /// - `1` - `payload` is exactly 20 bytes and contains an SHA-1 root hash.
+    /// - `2` - `payload` is exactly 32 bytes and contains an SHA-256 root hash.
+    /// To only revoke a claim without creating a new one, pass root hash `0`,
```
NunoAlexandre, line 19:
```suggestion
    /// To revoke a claim without creating a new one, pass payload `0`,
```

contracts/Claims.sol
```diff
@@ -0,0 +1,27 @@
...
+    event Claimed(address indexed addr);
+
+    /// @notice Creates a new claim of a Radicle Link identity.
+    /// Every new claim invalidates previous ones made with the same account.
+    /// The claims have no expiration date and don't need to be renewed.
+    /// If either `format` is unsupported or `payload` is malformed as per `format`,
+    /// the previous claim is revoked, but a new one isn't created.
+    /// Don't send a malformed transactions on purpose, to properly revoke a claim see `format`.
+    /// @param format The format of `payload`, currently supported values:
+    /// - `1` - `payload` is exactly 20 bytes and contains an SHA-1 root hash.
```
NunoAlexandre, line 17: regarding these references to "root hash", I wonder if we should just remove them or make it clear what it is referring to, since now there is no context to it.

cloudhead: Hmm yeah this doc will get outdated as soon as we extend the payload beyond those two things. On the other hand, the docs as-is isn't wrong: the expected payload _is_ the root hash when format = 1

NunoAlexandre: yap. maybe say "Radicle Identity root hash"?

CodeSandwich: I've changed it into:
```suggestion
    /// - `1` - `payload` is exactly 20 bytes and contains an SHA-1 Radicle Identity root hash
```

Is that clear enough?

contracts/Claims.sol
CodeSandwich: Fixed

NunoAlexandre (approved):
:speedboat:

PR #133, opened 2021-03-23
chore: drop ethregistrar dependency

NunoAlexandre:
It can not be built on node >10 due to [an issue with scrypt](https://github.com/barrysteyn/node-scrypt/pull/197), which makes integrating this latest revision on Upstream unviable. We only need the JSON contract files so I have tracked them and pointed to those instead, and dropped that dependency.

### DIFF
--- a/ci/base-image/Dockerfile
+++ b/ci/base-image/Dockerfile
@@ -1,1 +1,1 @@
-FROM node:10.23.1-buster
+FROM node:12.18.3-buster
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -3,7 +3,7 @@
   command: "ci/run"
   env:
     SHARED_MASTER_CACHE: true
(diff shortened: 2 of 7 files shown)


ci/base-image/Dockerfile
```diff
@@ -1 +1 @@
-FROM node:10.23.1-buster
+FROM node:12.18.3-buster
```
CodeSandwich, line 1: After changing of the docker image, don't we need to update `.buildkite/pipeline.yaml`? [DEVELOPING.md](https://github.com/radicle-dev/radicle-contracts/blob/master/DEVELOPING.md#updating-cis-base-docker-image)

CodeSandwich (changes requested):
Nice, heavy-handed but works :+1:

ci/base-image/Dockerfile
NunoAlexandre: :heavy_check_mark:

CodeSandwich (approved):

PR #138, opened 2021-04-30
More events

CodeSandwich:
Adds more pool events needed to switch from polling the contract state to listening to the events: https://github.com/orgs/radicle-dev/projects/18#card-54079531

### DIFF
--- a/contracts/Pool.sol
+++ b/contracts/Pool.sol
@@ -128,6 +128,13 @@ abstract contract Pool {
     /// The receiver will be getting `weight / PROXY_WEIGHTS_SUM` of funds sent to the proxy.
     event ProxyToReceiverUpdated(address indexed proxy, address indexed receiver, uint32 weight);
 
+    /// @notice Emitted when a sender is updated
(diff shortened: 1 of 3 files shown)


contracts/Pool.sol
```diff
@@ -128,6 +128,18 @@ abstract contract Pool {
     /// The receiver will be getting `weight / PROXY_WEIGHTS_SUM` of funds sent to the proxy.
     event ProxyToReceiverUpdated(address indexed proxy, address indexed receiver, uint32 weight);
 
+    /// @notice Emitted when a sender is updated
+    /// @param sender The updated sender
+    /// @param balance The sender's balance since the event block's timestamp
+    /// @param amtPerSec The target amount sent per second after the update.
+    /// Takes effect since the event block's timestamp (inclusively).
```
geigerzaehler, line 135:
```suggestion
    /// Takes effect on the event block's timestamp (inclusively).
```

geigerzaehler (approved):

contracts/Pool.sol

PR #139, opened 2021-05-07
Demote production dependencies to dev deps

geigerzaehler:
Both `@ensdomains/ens` and `@uniswap/v2-periphery` can by dev
dependencies. The packages are only used to build the project. The
build artifacts don’t contain references to the packages.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,11 +4,9 @@
   "license": "GPL-3.0-only",
   "main": "build/src/index.js",
   "types": "build/src/index.d.ts",
-  "dependencies": {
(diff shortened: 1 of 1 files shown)


CodeSandwich:
This split has been removed before and then reintroduced in https://github.com/radicle-dev/radicle-contracts/commit/e57fdeb4981208d0d5f621da3d818899131ee141, now we're reremoving it here. Are you sure that hiding these deps has no effect on the clients?

CodeSandwich (approved):
Good points!

PR #141, opened 2021-05-12
Update @typechain/ethers-v5 to have strict contract types

geigerzaehler:
The update to `@typechain/ethers-v5` includes https://github.com/ethereum-ts/TypeChain/pull/373. This ensures that the interfaces generated for contracts are typed and we can’t call arbitrary functions on contracts.

The update also requires us to update `typechain` package from `v4` to `v5`. This results in changes to `hardhat.config.ts`.

Finally, we add dependencies that are required by the generated contract bindings.


geigerzaehler:
See https://github.com/radicle-dev/radicle-upstream/pull/1888 for the integration PR.

CodeSandwich (approved):

CodeSandwich:
I was able to revert the dependencies back to the state where there are no non-dev deps and we don't touch the `@ethersproject/...` packages: https://github.com/radicle-dev/radicle-contracts/pull/143 The CI passes and it seems to be working fine as an upstream dependency. Am I missing something?

PR #143, opened 2021-05-13
Igor/update typechain ethers

CodeSandwich:


geigerzaehler:
On CI it does not matter whether something is in `dependencies` or `devDependencies` because `yarn install` will install all dependencies, including `devDependencies`. In addition, the `@ethersproject/*` dependencies are installed because they are themselves dependencies of `ethers`.

Although all this works on CI it causes problems when `radicle-contracts` is used as a dependency. Then, only the `dependencies` are installed. This means that if the dependent (in our case Upstream) does not declare `ethers` as a dependency then using `radicle-contracts` will throw an exception because a dependency is missing. In our case Upstream is using a dependency but we don’t want to put that obligation on the consumer. It may also be the case that the generated contracts require a minimum version of `ethers`. This also can’t be enforced if we don’t specify it as a production dependency.

My method for choosing `dependencies` is to include everything that is explicitly referenced in the contract bindings.

Sorry for the back-and-forth on this and the confusion. It also took my some time to wrap my head around this and think of all the possible problems. NPM just isn’t great with this.

CodeSandwich:
Thank you for the explanation! How can I reproduce that issue? I need to know how to check if I'm breaking something in upstream after tinkering around the radicle-contracts deps, otherwise sooner or later this bug will be reintroduced.

geigerzaehler:
> Thank you for the explanation! How can I reproduce that issue? I need to know how to check if I'm breaking something in upstream after tinkering around the radicle-contracts deps, otherwise sooner or later this bug will be reintroduced.

I only have some ideas how to check this programatically and don’t know about a definitive solution. One idea is to bundle library with Webpack (or another bundler) and tell it to only use `dependencies`. Another idea is to leverage Yarn v2 somehow. With it’s [PnP linker](https://yarnpkg.com/features/pnp) only production dependencies are available when it is run with `yarn node`. Not sure how to program this check though.

I’d also be ok with not developing a system to check this and just follow the guideline. Since Upstream is the only consumer we should be able to fix breaking updates quite easily.

PR #146, opened 2021-06-11
Document contract addresses

rudolfs:
Document latest deployments so they're easy to find.


.nvmrc
```diff
@@ -1 +0,0 @@
-v12
```
rudolfs, line 1: Removing this because it's outdated and prevented me from deploying.

CodeSandwich (approved):

geigerzaehler (approved):

PR #147, opened 2021-06-17
Add funding pool deployment scripts

CodeSandwich:


geigerzaehler (approved):

rudolfs (commented):
Looks good. Should we deploy these on Ropsten/Rinkeby?

CodeSandwich:
No, at least not yet, I've added it for my own usage and for completeness sake. It'll be great to move upstream to the DAI-specific funding pool though, it'll let us get rid of the explicit approval step.

PR #151, opened 2021-07-09
Copy declaration files clean-ups

CodeSandwich:
Changes proposal for https://github.com/radicle-dev/radicle-contracts/pull/150

We can clean up a lot of complexity and lint violations if we switch the script from JS to TS.


tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "include": [
     "src/**/*.ts",
     "test/**/*.ts",
-    "scripts/**/*.js",
+    "scripts/**/*.ts",
```
CodeSandwich, line 5: I'm not sure if it's even needed, ts-node seems to be working without need for `tsc`

package.json
```diff
@@ -37,7 +37,7 @@
     "web3": "^1.3.5"
   },
   "scripts": {
-    "build": "hardhat compile && tsc && node scripts/copy-contract-declaration-files.js",
+    "build": "hardhat compile && tsc && ts-node -e 'require(\"./scripts/copy-contract-declaration-files.ts\").main();'",
```
geigerzaehler, line 40: Couldn’t we just do `ts-node ./scripts/copy-contract-declaration-files.ts` and call `main()` there? This would make the call here easier to understand and you could call the script also from the commandline.

geigerzaehler (commented):
Thanks a lot! Having this written in typescript indeed makes more sense. I have just one question.

tsconfig.json
geigerzaehler: This is useful for type-checking though. If we were to omit this, running `tsc` wouldn’t catch any type errors.

package.json
CodeSandwich: We could, but that would require the awkward:
```ts
// eslint-disable-next-line @typescript-eslint/no-floating-promises
(async function main() { /*YOUR_CODE_HERE*/ })();
```
IMO it's bad, because it uses a hidden node environment feature to work around JS's lack of support for async code in module execution. It's hard for me as a human to understand and even eslint is alarmed. `require(...).main();` feels much cleaner.

geigerzaehler: Thanks for the explanation. We are calling `main()` in scripts [in radicle-upstream](https://github.com/radicle-dev/radicle-upstream/blob/master/scripts/license-header.ts#L164) and I’d say this approach is rather widespread in the JavaScript ecosystem. (They’re less hesitant there to go with less clean solutions :stuck_out_tongue:.) I’ll leave it up to you though.

CodeSandwich: Ok, that's a good point, let's do that. It turns out that for whatever reason `main().catch((err) => { throw err; });` works without annoying lint or tsc, which is great. I don't understand why it works or why any deviations like `await main()...` or plain `main();` don't, but hey, it does the job :shrug:

geigerzaehler (approved):

PR #152, opened 2021-07-12
Remove all code related to exchanges

geigerzaehler:
The registrar does not need an exchange since #94. This should make deployment faster.


CodeSandwich (approved):
Makes sense :+1:

CodeSandwich (approved):
Still makes sense :+1:

PR #154, opened 2021-07-26
Add more data to Claimed event

CodeSandwich:
A part of https://github.com/radicle-dev/radicle-upstream/issues/2158. Removes the need to read calldata to understand the claim.

### DIFF
--- a/contracts/ClaimsV2.sol
+++ b/contracts/ClaimsV2.sol
@@ -1,0 +1,27 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.7.5;
+
+// @notice Used to create Radicle Link identity claims
+contract ClaimsV2 {
+    /// @notice Emitted on every Radicle Link identity claim
+    /// @param addr The account address which made the claim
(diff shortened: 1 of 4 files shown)


geigerzaehler (commented):
How do you feel about duplicating this contract as `ClaimsV2` since the other contract is already deployed to mainnet?

CodeSandwich:
If we're going to keep using the old version, then yes, we definitely should keep its source code. But if we're going to abandon the old contract, then I don't think that its code is worth keeping in our repo.

CodeSandwich:
It seems like we may keep using the old contract, I've added a separate V2

geigerzaehler:
> If we're going to keep using the old version, then yes, we definitely should keep its source code. But if we're going to abandon the old contract, then I don't think that its code is worth keeping in our repo.

I think it would still make sense to have a separate contract: If we merge this PR now, then we’re forced to integrate the change to the contract in upstream when we want to update the dependency on this repo. I don’t think that would be ideal.

geigerzaehler (approved):

